PR #26, opened 2020-09-15
Improve primary interface

fmauch:
This branch will

* Add a full primary client object structure to the client
* Read more data from the primary interface
* Show users when they send malformed script code to the robot
* Inform users when they are trying to send script code to an e-Series robot which is not in remote control mode
* Output any error messages coming up on the robot

I am not sure yet, to which extent all of those things will be tackled in this MR, but it will definitely be a start from which such features could easily be implemented.


fmauch:
Closing this in favor of #186

Issue #28, opened 2020-11-22
I can't cmake ,please help me.


aoligeihahah:
/home/he/catkin_ws/src/Universal_Robots_Client_Library/include/ur_client_library/rtde/data_package.h:32:19: fatal error: variant:There is no such file or directory

aoligeihahah:
![2020-11-22 22-00-43屏幕截图](https://user-images.githubusercontent.com/72695459/99905819-3a8c5700-2d0e-11eb-89ff-333b414f3759.png)

fmauch:
Please give us a little bit more information:
 - On what kind of system are you on?
 - Which compiler are your using?
 - Which branch of this library are you trying to build?

Also please give the full build output using a [code block](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code), not a screenshot.

aoligeihahah:
sorry.
My system is:
ubantu16.04+ros-kinetic
My code is:
mkdir -p catkin_ws/src && cd catkin_ws
git clone https://github.com/UniversalRobots/Universal_Robots_Client_Library.git src/Universal_Robots_Client_Library
sudo apt update -qq
rosdep update
rosdep install --from-paths src --ignore-src -y
catkin_make_isolated
source devel_isolated/setup.bash

gavanderhoorn:
> My system is:
ubantu16.04+ros-kinetic

Does that support C++17?

aoligeihahah:
How to check whether the  my system supports

aoligeihahah:
> How to check whether the my system supports

ok，It should be the problem.I tried get the system to support re installation.

fmauch:
You can use the boost branch of the library that will not require c++17.

fmauch:
It would still be nice if you could post your full cmake output as in theory we have a check for c++17 in place that should result in a cmake error already.

aoligeihahah:
Thank you. I installed gcc-9.1.0, which solved this problem. But I had a problem. I succeeded in roslaunch example_ organization_ Ur_ Launch ex-ur10-1.launch, but catkin_ make_ Isolated has the error shown in the figure below. I want to get help.

my code：
  cd catkin_ ws && source devel/ setup.bash
  catkin_ make_ isolated

Create an example_organization_ur_launch : 
catkin_ create_ pkg example_ organization_ Ur_ launch ur_ client_ library

[...]

PR #29, opened 2020-11-24
Improve documentation

fmauch:
* Remove unstable disclamer
 * Clarify C++17 dependency

It occurred from #28 that the c++17 dependency isn't documented clearly enough.


t-schnell (approved):
Did not find any problems with the changed documentation.

PR #32, opened 2021-02-17
Add Isaac build

urmahp:
Used when building client library inside Isaac SDK.

### DIFF
--- a/include/ur_client_library/log.h
+++ b/include/ur_client_library/log.h
@@ -28,6 +28,11 @@
 #define LOG_ERROR CONSOLE_BRIDGE_logError
 #define LOG_FATAL CONSOLE_BRIDGE_logError
 
+#elif defined(ISAAC_BUILD)
+#include "engine/core/logger.hpp"
+
+#define LOG_WARN(format, ...)
+
 #else
 
 #define LOG_DEBUG(format, ...) printf("[DEBUG]: " format "\n", ##__VA_ARGS__)
--- a/include/ur_client_library/log.h
+++ b/include/ur_client_library/log.h
@@ -31,7 +31,7 @@
 #elif defined(ISAAC_BUILD)
 #include "engine/core/logger.hpp"
 
-#define LOG_WARN(format, ...)
+#define LOG_WARN(format, ...) LOG_WARNING(format, ##__VA_ARGS__)
 
 #else
 


include/ur_client_library/log.h
```diff
@@ -28,6 +28,11 @@
 #define LOG_ERROR CONSOLE_BRIDGE_logError
 #define LOG_FATAL CONSOLE_BRIDGE_logError
 
+#elif defined(ISAAC_BUILD)
+#include "engine/core/logger.hpp"
+
+#define LOG_WARN(format, ...)
```
fmauch, line 34: Why is there only a define for `LOG_WARN` and why is it empty?

fmauch (commented):
In general, it feels wrong to include this here. The library part should not have to be aware of any application's environments.

Currently, however, I don't know a better solution for this.

include/ur_client_library/log.h
urmahp: I have added a commit where I define LOG_WARN to use LOG_WARNING defined in "engine/core/logger.hpp"
The other defines are declared within "engine/core/logger.hpp" file.

urmahp:
These changes should be added to the boost branch as well, do I open a new PR on the boost branch, or what is the correct workflow for adding the same changes to the boost branch?

fmauch:
> These changes should be added to the boost branch as well, do I open a new PR on the boost branch, or what is the correct workflow for adding the same changes to the boost branch?

I would apply the same commit to the boost branch, as well. Please remind me, if I forget it.

include/ur_client_library/log.h
fmauch: That answers the second question. But what about the other log macros?

urmahp: LOG_DEBUG, LOG_INFO and LOG_ERROR are all defined inside "engine/core/logger.hpp", so it should be enough to just indclude "engine/core/logger.hpp".

LOG_FATAL is not defined inside the file, so I will fix that. 
The build wasn't failing, because LOG_FATAL isn't used anywhere in the library.

fmauch: And `engine/core/logger.hpp` is where?

It seems like this is requiring files from the ISAAC driver, which does not seem right.

urmahp: "engine/core/logger.hpp" is part of the Isaac SDK framework.

urrsk:
> In general, it feels wrong to include this here. The library part should not have to be aware of any application's environments.
> 
> Currently, however, I don't know a better solution for this.

I agree that we need to find a better solution in general for this.
Currently we are playing the the idea of making a simple public subscribe behavior, to move the definition of how to report a log message to the drivers side.

fmauch:
Closing this in favor of #40

PR #34, opened 2021-02-22
Beta testing

stefanscherzinger:
Provide current state for beta-testing also for the boost branch. This shall include users on ROS kinetic.


t-schnell (approved):
nothing new to say here

PR #36, opened 2021-02-22
Fixed links in README (#35)

stefanscherzinger:
After merging in the separation on the driver, those links got outdated.


stefanscherzinger:
Already fixed in #37

PR #41, opened 2021-03-10
Added LogHandler, this enables the possibility to change the behavior when logging messages with this library.

urmahp:
The loghandler changes from PR #40 needs to be applied to the boost branch as well.


fmauch:
I'll do a PR with the master changes separately

PR #47, opened 2021-03-18
Add function to set keepalive counter.

urmahp:
This enables the possibility to set the number of timeout reads on the robot. 

The functions to set the keepalive count should probably be changed a bit once #46 has been merged.

### DIFF
--- a/include/ur_client_library/comm/reverse_interface.h
+++ b/include/ur_client_library/comm/reverse_interface.h
@@ -62,7 +62,7 @@ public:
    *
    * \param port Port the Server is started on
    */
-  ReverseInterface(uint32_t port) : server_(port)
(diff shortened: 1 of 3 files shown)


fmauch:
I think, this is not needed anymore with the implementation from #46, right? So I will close this.

urmahp:
> I think, this is not needed anymore with the implementation from #46, right? So I will close this.

Isaac has a hard time cooping with the robots cycle time, meaning that sometimes connection to the robot will be lost, because of scoket timeout reads on the robot. This will enable the possibility to set a number of timeout reads before loosing connection to the robot. 
It needs to be updated in relation to the changes in #46, so I will update and reopen this PR once the other PR has been merged.

PR #56, opened 2021-04-23
Use colcon as builder

fmauch:
This should also run plain ctests


fmauch:
Closing this PR. Noetic uses colcon, anyway.

PR #57, opened 2021-04-28
Install package XML for ROS2 installs

puck-fzi:
This PR installs the package.xml for ROS2 builds.

This is similar to the REP 137 for catkin, since a corresponding REP for colcon is currently missing.

For the package.xml condition attribute the package.xml format has to be bumped to version 3 ( as specified in REP 149)


fmauch (approved):
LGTM.

PR #62, opened 2021-05-06
Install the resources folder instead of the script file directly

fmauch:
The main problem before was that the resource file got installed into another
relative package location (/) than in the source package (/resources).
This way, other packages searching that file would not work with both, source
builds and installed versions.


fmauch:
Merging although noetic pipeline is failing in downstream workspace setup. Melodic and foxy are running fine.

PR #63, opened 2021-05-06
Use file= fields for license tags in package.xml

fmauch:
Since we have multiple licenses of different components, it might make sense to hint users to the respective license files inside the package.xml

Note: This was raised in https://github.com/ros/rosdistro/pull/29451


package.xml
```diff
@@ -10,10 +10,10 @@
   <author>Lea Steffen</author>
   <author>Tristan Schnell</author>
   <maintainer email="exner@fzi.de">Felix Exner</maintainer>
-  <license>Apache-2.0</license>
-  <license>BSD-2-Clause</license>
-  <license>Zlib</license>
-  <license>MIT</license>
+  <license file="LICENSE_apache_2_0">Apache-2.0</license>
```
nuclearsandwich, line 13:
```suggestion
  <license file="LICENSE">Apache-2.0</license>
```

Shouldn't this be simply `LICENSE` to match the filename in the repository. As far as I can tell there is no `LICENSE_apache_2_0` file in this repository.

nuclearsandwich (commented):
This is exactly what I was thinking. Thanks for the quick turnaround.

package.xml
fmauch: I renamed it also in this PR. But I wasn't unsure which would be better.

nuclearsandwich: Ah, I somehow missed the file rename. In my experience a lot of automation searches for and assumes a `LICENSE` or `COPYING` file at the repository root. The only reason _not_ to provide that file is to avoid giving the impression that there are no other licenses present in the repository to those tools. But since the ROS tooling uses your package.xml to assess license files I'm happy either way!

fmauch: It was indeed my intention to emphasize the fact that this is not the only license file present in this repo.

fmauch:
Merging this as no objections came up.

PR #64, opened 2021-05-06
Do not fail fast in actions

fmauch:
If there's a problem in one ROS distro, we still want to see whether the others are building correctly


fmauch:
Merging although the noetic build is currently failing on setting up the downstream workspace. That was exactly the reason I made this PR in the first place.

PR #65, opened 2021-05-10
Calibration check optional

fmauch:
As not all downstream packages support extracting calibration information from the robot we make this check optional. This moves calling the calibration check towards the application layer.

@urmahp This should remove the error output from the ISAAC driver.

@urrsk FYI, as we've been talking about that.

### DIFF
--- a/include/ur_client_library/ur/calibration_checker.h
+++ b/include/ur_client_library/ur/calibration_checker.h
@@ -95,9 +95,21 @@ public:
     return checked_;
   }
 
+  /*!
+   * \brief Returns whether the calibration check was successful.
+   *
+   * \returns True if the robot's calibration checksum matches the one given to the checker. False
+   * if it doesn't match or the check was not yet performed.
+   */
+  bool checkSuccessful()
+  {
+    return matches_;
+  }
+
 private:
   std::string expected_hash_;
   bool checked_;
+  bool matches_;
 };
 }  // namespace urcl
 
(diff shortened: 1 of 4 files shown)


urmahp:
When testing this with the ROS driver, it is working very well. 

When I test it with the Isaac driver, i get an error saying that the call to the constructor without the calibration checksum is ambiguous. I suspect that it might be because the constructor with the checksum, has a default value for the calibration checksum. But then I would expect it to happen for the ROS driver as well. I will try to investigate this further on friday.

fmauch:
> When testing this with the ROS driver, it is working very well.
> 
> When I test it with the Isaac driver, i get an error saying that the call to the constructor without the calibration checksum is ambiguous. I suspect that it might be because the constructor with the checksum, has a default value for the calibration checksum. But then I would expect it to happen for the ROS driver as well. I will try to investigate this further on friday.

The ROS driver still uses the constructor **with** the calibration checksum for backwards compatibility, that's why it isn't happening there.

It would probably make sense to remove the default value, as it isn't needed anymore with that.

urmahp:
> The ROS driver still uses the constructor **with** the calibration checksum for backwards compatibility, that's why it isn't happening there.

When testing it locally I updated the ROS driver to use the constructor without the checksum, but I will investigate it further today.

urmahp:
I found the issue, it is in the Isaac driver and not in the library. 

Currently NULL is inserted as the tool communication setup parameter in the constructor in the Isaac driver. This means that this [constructor](https://github.com/fmauch/Universal_Robots_Client_Library/blob/calibration_check_optional/include/ur_client_library/ur/ur_driver.h#L86) and this [constructor](https://github.com/fmauch/Universal_Robots_Client_Library/blob/calibration_check_optional/include/ur_client_library/ur/ur_driver.h#L140) are ambiguous, because NULL could both be the calibration string or the tool communication setup. Instead of NULL I updated it to use an empty ToolCommSetup pointer instead, which fixes the issue. 

So the update removes the calibration error, and it is working with both the Isaac driver and the ROS driver.

fmauch:
thanks for testing @urmahp

PR #68, opened 2021-06-01
Added acknowledgment section with partner logos

fmauch:
as per subject


urrsk (approved):

README.md
```diff
@@ -383,3 +387,31 @@ application, as by default `console_bridge` will only print messages of level WA
 See [`examples/primary_pipeline.cpp`](examples/primary_pipeline.cpp) as an example.
 
 The ROS logger will be moved to the ROS driver in a future release.
+
+
+## Acknowledgment
+Many parts of this library are forked from the [ur_modern_driver](https://github.com/ros-industrial/ur_modern_driver).
+
+Developed in collaboration between:
+
```
urrsk, line 396: Why this newline?

urrsk: I think it makes sense to have a newline. But I would be nice if it was done in the same way on all the repos

fmauch: Done in the other repos, as well. Thanks for spotting this!

PR #70, opened 2021-06-01
Refactor reverse interface

fmauch:
This moves the ReverseInteface and the ScriptSender from the `comm` namespace to the `control` namespace. As we plan to extend the control protocol structure with the changes from the beta-testing branch, I think it makes sense to clean up things a little beforehand.

Note: I decided to keep the `ControlMode` enums inside the `comm` namespace first because of backwards compatibility, as they are used on application level, and second because they actually define the communication protocol and therefore seem to be not too wrong inside the comm namespace.

### DIFF
--- a/src/control/reverse_interface.cpp
+++ b/src/control/reverse_interface.cpp
@@ -43,7 +43,7 @@ ReverseInterface::ReverseInterface(uint32_t port, std::function<void(bool)> hand
   server_.start();
 }
 
-bool ReverseInterface::write(const vector6d_t* positions, const ControlMode control_mode)
+bool ReverseInterface::write(const vector6d_t* positions, const comm::ControlMode control_mode)
 {
   if (client_fd_ == -1)
   {
--- a/src/control/reverse_interface.cpp
+++ b/src/control/reverse_interface.cpp
@@ -1,0 +1,109 @@
+// this is for emacs file handling -*- mode: c++; indent-tabs-mode: nil -*-
+
+// -- BEGIN LICENSE BLOCK ----------------------------------------------
+// Copyright 2021 FZI Forschungszentrum Informatik
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


src/control/reverse_interface.cpp
```diff
@@ -0,0 +1,109 @@
...
+void ReverseInterface::disconnectionCallback(const int filedescriptor)
+{
+  URCL_LOG_INFO("Connection to reverse interface dropped.", filedescriptor);
+  client_fd_ = -1;
+  handle_program_state_(false);
+}
+
+void ReverseInterface::messageCallback(const int filedescriptor, char* buffer)
+{
+  URCL_LOG_WARN("Messge on ReverseInterface received. The reverse interface currently does not support any message "
```
stefanscherzinger, line 105:
```suggestion
  URCL_LOG_WARN("Message on ReverseInterface received. The reverse interface currently does not support any message "
```

src/control/script_sender.cpp
```diff
@@ -0,0 +1,80 @@
...
+  server_.setMessageCallback(
+      std::bind(&ScriptSender::messageCallback, this, std::placeholders::_1, std::placeholders::_2));
+  server_.setConnectCallback(std::bind(&ScriptSender::connectionCallback, this, std::placeholders::_1));
+  server_.setDisconnectCallback(std::bind(&ScriptSender::disconnectionCallback, this, std::placeholders::_1));
+  server_.start();
+}
+
+void ScriptSender::connectionCallback(const int filedescriptor)
+{
+  URCL_LOG_DEBUG("New client connected at FD %d.", filedescriptor);
```
stefanscherzinger, line 46:
```suggestion
  URCL_LOG_DEBUG("New client connected at file descriptor %d.", filedescriptor);
```
Is _FD_ sufficiently known as a term? If yes, then it's ok.

src/control/script_sender.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void ScriptSender::disconnectionCallback(const int filedescriptor)
+{
+  URCL_LOG_DEBUG("Client at FD %d disconnected.", filedescriptor);
```
stefanscherzinger, line 51:
```suggestion
  URCL_LOG_DEBUG("Client at file descriptor %d disconnected.", filedescriptor);
```

stefanscherzinger (approved):
I read over this and it looks fine.

src/control/script_sender.cpp
fmauch: It's a debug output anyway...

PR #74, opened 2021-06-09
Removed console_bridge dependency

urmahp:
As log handlers for the client library has been created in the drivers, the console bridge dependency is no longer needed.

As a log handler has been created for the ROS2 Driver and the ROS Driver, the console bridge dependency is no longer needed. Also in order for the handlers to actually work, we need to build the client library without console bridge. 

This shouldn't be merged before the two PR's https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/398 and https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/126 has been merged

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,18 +60,6 @@ if(CMAKE_THREAD_LIBS_INIT)
   target_link_libraries(urcl PUBLIC "${CMAKE_THREAD_LIBS_INIT}")
 endif()
 
-find_package(console_bridge)
-if(console_bridge_FOUND)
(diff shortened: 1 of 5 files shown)


fmauch:
At a first glimpse, this looks good. I'll postpone this until after the next release containing all the Cartesian features, alright?

fmauch (approved):
Looks good to me! Thanks @urmahp

fmauch:
As both mentioned PRs are merged, I would go ahead and merge this.

PR #78, opened 2021-06-21
Always install package.xml file

fmauch:
As this package's build type is plain cmake, catkin is not necessarily
available at build time. Hence, checking for catkin before installing the
package.xml doesn't make any sense as this will never work for the buildfarm builds.

Installig the package.xml file in every case won't hurt in non-ROS environments.

When the package.xml file is not installed, toolings like rospack will not be
able to find the package.

@puck-fzi @urmahp opinions on this?

### DIFF
--- a/.github/workflows/industrial-ci.yml
+++ b/.github/workflows/industrial-ci.yml
@@ -30,6 +30,7 @@ jobs:
             ROS_REPO: main
             IMMEDIATE_TEST_OUTPUT: true
             DOWNSTREAM_WORKSPACE: "github:UniversalRobots/Universal_Robots_ROS_Driver#master https://raw.githubusercontent.com/UniversalRobots/Universal_Robots_ROS_Driver/master/.noetic.rosinstall"
+            BUILDER: catkin_tools
             DOCKER_RUN_OPTS: --network static_test_net
             BEFORE_INIT: 'apt-get update -qq && apt-get install -y iproute2 iputils-ping && ip addr && ping -c5 192.168.56.101'
             URSIM_VERSION: 5.8.0.10253
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,11 +104,9 @@ write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/ur_client_libraryCo
 install(FILES ur_client_libraryConfig.cmake
   DESTINATION lib/cmake/ur_client_library)
 
-# When built with catkin available, install package.xml
-find_package(catkin QUIET)
-find_package(ament_cmake QUIET)
-if(catkin_FOUND OR ament_cmake_FOUND)
-  install(FILES package.xml DESTINATION share/${PROJECT_NAME})
-endif()
+# Install package.xml file so this package can be processed by ROS toolings
+# See REP 136 for details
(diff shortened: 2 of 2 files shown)


puck-fzi:
LGTM.

I don't see any problems with always installing. 
The check for catkin or ament_cmake seemed more like a workaround to see if the system already supports ROS or ROS2. But not necessarily if the lib should be used in that context anyway

fmauch:
Thanks for looking at this @gavanderhoorn, @puck-fzi 

I just don't quite understand why the noetic builds fail with this change, while they succeed with the master branch atm.

gavanderhoorn:
What about registering the package in the ROS 2 ament/resource index?

gavanderhoorn:
> From what I understand the docs this would be an alternative to installing the resource folder to the package's share location as we currently do.

I'm not sure I understand. The ament index is basically a registry containing information on which locations should be considered when looking for certain resources. This was introduced to reduce the amount of fs crawling which is the default in ROS 1 (by `rospkg` fi).

For packages to be 'findable' by ROS 2 tooling, registration in the index is required.

The rationale for doing that with the package in this repository would seem to be similar to what you're doing in this PR (ie: to be able to offer resources which can be located relative to the package, instead of by absolute paths).

> However, I assume that we should still install the `package.xml` file, so this won't be a true alternative to this PR, right?

it's not an alternative to anything :)

It's similar in purpose (see above).

If the client library is purely a CMake project and you don't want to offer resources which should be locatable relative to a "ROS package" (whether ROS 1 or ROS 2), then there's no point in registering. But seeing as you're going to install your manifest, I believe you do want to do that.

fmauch:
> Note that "registration" basically comes down to placing a marker file in a certain location (like [this](https://github.com/ros-industrial/ros_industrial_cmake_boilerplate/blob/baa0e400946ddf5dc7458559430a0d57bac0a50d/ros_industrial_cmake_boilerplate/cmake/cmake_tools.cmake#L135-L144)).

Thanks for the elaboration.

fmauch:
After spending some time debugging this, the problem seems that as soon as `find_package(catkin QUIET)` is removed from the `CMakeLists.txt` file rosdep in the downstream workspace seems to not recognize this package anymore (if being built with colcon) and instead installs the released binary apt package. Then, there seem to be shadowing issues with the installed binary package...

When changing the builder to `catkin_tools` the released version doesn't get installed and everything works as expected.

I am unsure how to proceed at this point...

gavanderhoorn:
> After spending some time debugging this, the problem seems that as soon as `find_package(catkin QUIET)` is removed from the `CMakeLists.txt` file rosdep in the downstream workspace seems to not recognize this package anymore (if being built with colcon) and instead installs the released binary apt package. Then, there seem to be shadowing issues with the installed binary package...

on systems with ROS 2, that's because the package is not registered with the ament index, which is what `rosdep` uses there.

See also the lines in the CMake boilerplate I linked earlier.

fmauch:
To bring this forward, I will set the builder for noetic to catkin tools, as this error seems to only occur when using an underlay workspace that has been built using colcon.

Building against the released package or against the package being in the same workspace or using an underlay workspace built with catkin tools seems to work.

PR #85, opened 2021-08-05
Added target_frequency to RTDEClient

jornb:
For some scenarios, e.g. state monitoring, 500 Hz is overkill, and may lead to performance issues on some platforms or with multiple RTDE clients.

This PR adds a `target_frequency` argument to the `RTDE` constructor, defaulting to the maximum frequency (i.e. 125 or 500).

### DIFF
--- a/src/rtde/rtde_client.cpp
+++ b/src/rtde/rtde_client.cpp
@@ -34,7 +34,7 @@ namespace urcl
 namespace rtde_interface
 {
 RTDEClient::RTDEClient(std::string robot_ip, comm::INotifier& notifier, const std::string& output_recipe_file,
-                       const std::string& input_recipe_file)
+                       const std::string& input_recipe_file, double target_frequency)
   : stream_(robot_ip, UR_RTDE_PORT)
   , output_recipe_(readRecipe(output_recipe_file))
   , input_recipe_(readRecipe(input_recipe_file))
@@ -43,6 +43,7 @@ RTDEClient::RTDEClient(std::string robot_ip, comm::INotifier& notifier, const st
   , pipeline_(prod_, PIPELINE_NAME, notifier)
   , writer_(&stream_, input_recipe_)
   , max_frequency_(URE_MAX_FREQUENCY)
+  , target_frequency_(target_frequency)
 {
 }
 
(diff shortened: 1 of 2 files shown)


src/rtde/rtde_client.cpp
```diff
@@ -43,6 +43,9 @@ RTDEClient::RTDEClient(std::string robot_ip, comm::INotifier& notifier, const st
   , pipeline_(prod_, PIPELINE_NAME, notifier)
   , writer_(&stream_, input_recipe_)
   , max_frequency_(URE_MAX_FREQUENCY)
+  , target_frequency_(target_frequency)
+{
+}
```
fmauch, line 48: This seems wrong :-)

src/rtde/rtde_client.cpp
```diff
@@ -77,6 +80,10 @@ bool RTDEClient::init()
     max_frequency_ = CB3_MAX_FREQUENCY;
   }
 
+  // Cap target frequency, defaulting to maximum
+  if (target_frequency_ <= 0.0 || target_frequency_ > max_frequency_)
+      target_frequency_ = max_frequency_;
```
fmauch, line 85: We use braces in all (at least most) places.

```suggestion
      {
        target_frequency_ = max_frequency_;
      }
```

src/rtde/rtde_client.cpp
```diff
@@ -77,6 +80,10 @@ bool RTDEClient::init()
     max_frequency_ = CB3_MAX_FREQUENCY;
   }
 
+  // Cap target frequency, defaulting to maximum
+  if (target_frequency_ <= 0.0 || target_frequency_ > max_frequency_)
```
fmauch, line 84: I think, a thrown exception on an illegal frequency would make sense instead of silently going to the max value.

fmauch (commented):
Thank you for your contribution!

I added some minor comments, but I think there's also one major thing that could be changed: The `max_frequency_` member doesn't make much sense anymore. It could at least be a local variable instead.

src/rtde/rtde_client.cpp
jornb: I'll fix :)

src/rtde/rtde_client.cpp
jornb: OK!

src/rtde/rtde_client.cpp
jornb: I agree.

jornb:
Updated now based on previous review by @fmauch.

I agree with the `max_frequency_` field. However, I saw that it was used in `RTDEClient::getMaxFrequency()`, so I left it there, since it is not determined until call to `RTDEClient::init()`.

fmauch (approved):
LGTM! Thanks @jornb 

@stefanscherzinger if you agree, please go ahead and merge this.

PR #87, opened 2021-08-16
Merge Boost staging

fmauch:
Merge in latest developments from master via boost-staging


stefanscherzinger (approved):
Adding credits to relevant files looks good. I didn't check if you got all of them, though (hope so). Removing `console_bridge` and left-overs in docs looks also good.

Issue #94, opened 2022-02-08
startup before booted robot does not work


fmauch:
Starting up the client library interfaces before the robot is booted leads to multiple problems. This has been reported in the ROS driver, e.g. in https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/272

Basically, we have to distinguish two cases:

If the robot is not connected / the IP address is configured wrong, opening a connection to the remote socket fails immediately [here](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/8f20b42205f645bba406d5cf3b1163ea0a42fbc9/src/comm/tcp_socket.cpp?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L87) leading to [this exception](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/8f20b42205f645bba406d5cf3b1163ea0a42fbc9/include/ur_client_library/comm/producer.h?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L71). Mitigating this would require having a retry mechanism either in the TCPSocket or inside the producer which would effectively block the startup process until a connection can be established.

When the robot is actually connected, but not fully booted, the connection described above works. The primary interface connection seems to be coming up and seems to be usable afterwards (did a quick test using the primary_interface example and a URSim docker container). The RTDE-Interface, however, seems to have more problems. As far as I know the RTDE-Interface on the robot side is restarted at some point. So, if we are too early, the restart either happens in the middle of the handshake (I had a test run, where the client was waiting for confirming the configured communication frequency) or it the handshake is performed, but then the client is waiting for new data which it will never get, as the restarted RTDE server never made a proper handshake with the driver. This would effectively require the RTDEClient to be able to go back to an uninitialized state either when no new data is received or when the handshaking process is not behaving as expected (timeouting at some point). In that case we want to restart the handshake automatically (with accepting that we might have to wait for an answer for an undefined time with the first handshake step).

I hope that helps someone implementing this, maybe @urmahp you want to have a go at this?

urmahp:
[...]

PR #98, opened 2022-04-01
Support starting the driver, before the robot is booted

urmahp:
Refactored client library to support starting the client, before the robot is booted.

Changed producer to wait until robot is reachable on configured IP, instead of throwing an exception.

This will fix issue #94.

### DIFF
--- a/include/ur_client_library/comm/producer.h
+++ b/include/ur_client_library/comm/producer.h
@@ -66,9 +66,14 @@ public:
     tv.tv_sec = 1;
     tv.tv_usec = 0;
     stream_.setReceiveTimeout(tv);
-    if (!stream_.connect())
+    while (!stream_.connect())
     {
-      throw UrException("Failed to connect to robot. Please check if the robot is booted and connected.");
+      std::stringstream ss;
+      ss << "Failed to connect to robot on IP " << stream_.getHost()
+         << ". Please check that the robot is booted and reachable on " << stream_.getHost()
+         << ". Retrying in 10 seconds";
+      URCL_LOG_ERROR("%s", ss.str().c_str());
+      std::this_thread::sleep_for(std::chrono::seconds(10));
     }
   }
   /*!
--- a/include/ur_client_library/comm/producer.h
+++ b/include/ur_client_library/comm/producer.h
@@ -66,14 +66,9 @@ public:
     tv.tv_sec = 1;
     tv.tv_usec = 0;
     stream_.setReceiveTimeout(tv);
-    while (!stream_.connect())
+    if (!stream_.connect())
     {
-      std::stringstream ss;
-      ss << "Failed to connect to robot on IP " << stream_.getHost()
-         << ". Please check that the robot is booted and reachable on " << stream_.getHost()
-         << ". Retrying in 10 seconds";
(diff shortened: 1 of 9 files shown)


fmauch:
This will not completely fix #94 as we also have the connection [to the secondary interface](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/8f20b42205f645bba406d5cf3b1163ea0a42fbc9/src/ur/ur_driver.cpp#L72) and [to the dashboard server](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/8f20b42205f645bba406d5cf3b1163ea0a42fbc9/src/ur/dashboard_client.cpp#L48) would need a reconnect to fully support starting the driver before a booted robot.

src/rtde/rtde_client.cpp
```diff
@@ -44,22 +45,49 @@ RTDEClient::RTDEClient(std::string robot_ip, comm::INotifier& notifier, const st
...
+{
+  client_state_ = ClientState::INITIALIZING;
   // A running pipeline is needed inside setup
   pipeline_.init();
   pipeline_.run();
 
   uint16_t protocol_version = MAX_RTDE_PROTOCOL_VERSION;
-  while (!negotiateProtocolVersion(protocol_version))
+  // Protocol version should always be 1 when starting negotiation
+  parser_.setProtocolVersion(1);
```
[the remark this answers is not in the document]
urmahp: We should reset it every time before we start negotiating for it to be done properly, but I will move it to inside negotiateProtocolVersion.

urmahp:
> This will not completely fix #94 as we also have the connection [to the secondary interface](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/8f20b42205f645bba406d5cf3b1163ea0a42fbc9/src/ur/ur_driver.cpp#L72) and [to the dashboard server](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/8f20b42205f645bba406d5cf3b1163ea0a42fbc9/src/ur/dashboard_client.cpp#L48) would need a reconnect to fully support starting the driver before a booted robot.
> 
> @urmahp would you like to address this in this PR, as well?

If it can wait until next week, then I can have a look at it next week and address it in this PR.

fmauch:
> If it can wait until next week, then I can have a look at it next week and address it in this PR.

I would be fine with that. Maybe it will itch me enough before, to implement it myself, though ;-)

include/ur_client_library/comm/producer.h
```diff
@@ -68,7 +68,7 @@ class URProducer : public IProducer<T>
     stream_.setReceiveTimeout(tv);
     if (!stream_.connect())
     {
-      throw UrException("Failed to connect to robot. Please check if the robot is booted and connected.");
+      throw UrException("Failed to connect to robot. Please check if the robot is booted and connected");
```
fmauch, line 71: This change seems rather unnecessary.

fmauch (approved):
Looks good to me. I'll quickly add the missing dot in `producer.h` then we should be good to go.

include/ur_client_library/comm/producer.h
fmauch: resolved

PR #107, opened 2022-08-02
Fix Ros2 integration tests

fmauch:
Fixes integration tests for ROS2 distros


urmahp (approved):
LGTM and fixes the pipelines.

PR #108, opened 2022-08-02
Ignore atomicops.h in check_links

fmauch:
Since this is a 3rd-party file I don't really see any point in going deeper
into that.


fmauch:
The links that are still invalid are the ones fixed inside #107

fmauch:
Closing in favor of #117

PR #110, opened 2022-08-12
Added tcp_offset

urmahp:
The function tcp_offset isn't available before software version 3.12.0 and 5.5.1. It is needed in the ROS driver to transform the the force/torque measurements to the tcp frame.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,11 @@ implemented in order to create external applications leveraging the versatility 
 robotic manipulators.
 
 ## Requirements
+ * **Polyscope** (The software running on the robot controller) version **3.12.0** (for CB3-Series),
+   or **5.5.1** (for e-Series) or higher. If you use an older Polyscope version it is suggested to
+   update your robot. If for some reason (please tell us in the issues why) you cannot upgrade your
+   robot, please see the [version compatibility table](doc/polyscope_compatibility.md) for a
+   compatible tag.
  * The library requires an implementation of **POSIX threads** such as the `pthread` library
  * Socket communication is currently based on Linux sockets. Thus, this library will require Linux
    for building and using.
--- a/doc/polyscope_compatibility.md
+++ b/doc/polyscope_compatibility.md
@@ -1,0 +1,17 @@
+# Polyscope version compatibility
+The table below shows breaking changes in the library compared to Polyscope versions. Compatibility
+is listed for CB3 robots (versions 3.x.y) and e-Series robots (versions 5.x.y) respectively.
+
+It might be possible to use the library in a more recent version than the compatible versions listed
+below, but that might either require some manual modifications or not using a subset of features.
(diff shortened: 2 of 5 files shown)


fmauch (commented):
I cannot find `tcp_offset` inside the [RTDE guide](https://www.universal-robots.com/articles/ur/interface-communication/real-time-data-exchange-rtde-guide/). 

Since you didn't modify the recipe file: Where is that supposed to come from?

urmahp:
> I cannot find `tcp_offset` inside the [RTDE guide](https://www.universal-robots.com/articles/ur/interface-communication/real-time-data-exchange-rtde-guide/).
> 
> Since you didn't modify the recipe file: Where is that supposed to come from?

It is correct that the function isn't documented in the RTDE guide, but it is still part of the interface. 

I have modified the recipe file in the ROS driver, but I will go ahead and do it in the examples and the test in this repo as well.

fmauch:
Should we add a version requirement to the README or maybe even have a version check inside the interfaces where we can do that? Or would that be out of scope for this PR. Since we're raising the required Polyscope version with this PR, I think it makes sense to include something to the README at least.

urmahp:
> Should we add a version requirement to the README or maybe even have a version check inside the interfaces where we can do that? Or would that be out of scope for this PR. Since we're raising the required Polyscope version with this PR, I think it makes sense to include something to the README at least.

I definitely agree that we should update the README with the required polyscope versions.

I think that it would be a good idea to have some kind of version check, since we might add more features in the future that is not supported by older versions of polyscope. Since the version is already available in the rtde client it shouldn't be to much of a hassle to add it to this PR. 

There is probably a bit more work to figure out how to handle older versions in the ROS driver, since the function publishing the wrench is highly dependent on this function, at least for the e-series robots.

fmauch:
The version would not only affect the RTDE interface, but also primary and dashboard interfaces. Since all of them can be used standalone from this library, version checks would be needed everywhere.

fmauch:
[a5e177b](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/110/commits/a5e177b384e41f36f3d2c3e36f65bdfdecf76e3d) would be my suggestion for a minimal version compatibility notice (The tag does not yet exist, it would point to the last commit before this branch is merged). @urmahp @urrsk what do you think about that? This should give us the flexibility to announce such changes in the future, as well.

urrsk (approved):

urrsk:
> [a5e177b](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/110/commits/a5e177b384e41f36f3d2c3e36f65bdfdecf76e3d) would be my suggestion for a minimal version compatibility notice (The tag does not yet exist, it would point to the last commit before this branch is merged). @urmahp @urrsk what do you think about that? This should give us the flexibility to announce such changes in the future, as well.

Yes that look good and gives transparency

PR #111, opened 2022-08-16
Added interface for forwarding script commands to the robot, that is …

urmahp:
…then executed locally on the robot

This allows the user to be able to zero ftsensor, set payload and set tool voltage when the robot is in local control.

### DIFF
--- a/include/ur_client_library/control/script_command_interface.h
+++ b/include/ur_client_library/control/script_command_interface.h
@@ -37,17 +37,6 @@ namespace urcl
 namespace control
 {
 /*!
- * \brief Available script commands
- */
-enum class ScriptCommand : int32_t
-{
-
-  ZERO_FTSENSOR = 0,     ///< Zero force torque sensor
-  SET_PAYLOAD = 1,       ///< Set payload
-  SET_TOOL_VOLTAGE = 2,  ///< Set tool voltage
-};
-
-/*!
  * \brief The ScriptCommandInterface class starts a TCPServer for a robot to connect to and this connection is then used
  * to forward script commands to the robot, which will be executed locally on the robot.
  *
@@ -104,6 +93,17 @@ protected:
   virtual void messageCallback(const int filedescriptor, char* buffer, int nbytesrecv) override;
 
 private:
+  /*!
+   * \brief Available script commands
+   */
+  enum class ScriptCommand : int32_t
+  {
+
+    ZERO_FTSENSOR = 0,     ///< Zero force torque sensor
+    SET_PAYLOAD = 1,       ///< Set payload
+    SET_TOOL_VOLTAGE = 2,  ///< Set tool voltage
+  };
+
   bool client_connected_;
   static const int MAX_MESSAGE_LENGTH = 5;
 };
--- a/include/ur_client_library/control/script_command_interface.h
+++ b/include/ur_client_library/control/script_command_interface.h
@@ -1,0 +1,114 @@
+// this is for emacs file handling -*- mode: c++; indent-tabs-mode: nil -*-
+
+// -- BEGIN LICENSE BLOCK ----------------------------------------------
+// Copyright 2021 FZI Forschungszentrum Informatik
+// Created on behalf of Universal Robots A/S
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// -- END LICENSE BLOCK ------------------------------------------------
+
+//----------------------------------------------------------------------
+/*!\file
+ *
+ * \author  Mads Holm Peters mahp@universal-robots.com
+ * \date    2022-08-12
+ *
+ */
+//----------------------------------------------------------------------
+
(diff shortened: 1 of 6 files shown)


resources/external_control.urscript
```diff
@@ -213,13 +218,34 @@ def set_servo_pose(pose):
...
+thread script_commands():
+  while control_mode > MODE_STOPPED:
+    raw_command = socket_read_binary_integer(SCRIPT_COMMAND_DATA_DIMENSION, "script_command_socket", 0)
+    if raw_command[0] > 0:
+      command = raw_command[1]
+      if command == ZERO_FTSENSOR:
+        zero_ftsensor()
+      elif command == SET_PAYLOAD:
+        mass = raw_command[2] / MULT_jointstate
+        cog = [raw_command[3] / MULT_jointstate,raw_command[4] / MULT_jointstate,raw_command[5] / MULT_jointstate]
```
fmauch, line 231: Minor syntax nagging
```suggestion
        cog = [raw_command[3] / MULT_jointstate, raw_command[4] / MULT_jointstate, raw_command[5] / MULT_jointstate]
```

src/ur/ur_driver.cpp
```diff
@@ -223,6 +232,78 @@ bool UrDriver::writeTrajectoryControlMessage(const control::TrajectoryControlMes
...
+  }
+  else
+  {
+    if (script_command_interface_->clientConnected())
+    {
+      return script_command_interface_->zeroFTSensor();
+    }
+    else
+    {
+      std::stringstream cmd;
```
fmauch, line 256: Maybe a warning output about falling back to a secondary program might be useful here?

src/ur/ur_driver.cpp
```diff
@@ -223,6 +232,78 @@ bool UrDriver::writeTrajectoryControlMessage(const control::TrajectoryControlMes
...
+}
+
+bool UrDriver::setPayload(const float mass, const vector3d_t& cog)
+{
+  if (script_command_interface_->clientConnected())
+  {
+    return script_command_interface_->setPayload(mass, &cog);
+  }
+  else
+  {
```
fmauch, line 270: Same as above: Warning because of fallback might be good.

src/ur/ur_driver.cpp
```diff
@@ -223,6 +232,78 @@ bool UrDriver::writeTrajectoryControlMessage(const control::TrajectoryControlMes
...
+    case ToolVoltage::OFF:
+      break;
+    case ToolVoltage::_12V:
+      break;
+    case ToolVoltage::_24V:
+      break;
+    default:
+      std::stringstream ss;
+      ss << "The tool voltage should be 0, 12 or 24. The tool voltage is " << toUnderlying(voltage);
+      throw UrException(ss.str().c_str());
```
fmauch, line 296: I'm not remembering completely, whether we catch all exception types inside the driver. Having an `invalid_argument`  Exception seems more appropriate to me, but in general: It might not necessarily be the best choice to throw an exception here? We are inside a boolean function, so simply printing an error and returning `false` might also be an appropriate way of handling this, don't you think?

src/ur/ur_driver.cpp
```diff
@@ -223,6 +232,78 @@ bool UrDriver::writeTrajectoryControlMessage(const control::TrajectoryControlMes
...
+  }
+
+  if (script_command_interface_->clientConnected())
+  {
+    return script_command_interface_->setToolVoltage(voltage);
+  }
+  else
+  {
+    std::stringstream cmd;
```
fmauch, line 307: Again, warning about fallback would be nice.

include/ur_client_library/control/script_command_interface.h
```diff
@@ -0,0 +1,114 @@
...
+/*!
+ * \brief Available script commands
+ */
+enum class ScriptCommand : int32_t
+{
+
+  ZERO_FTSENSOR = 0,     ///< Zero force torque sensor
+  SET_PAYLOAD = 1,       ///< Set payload
+  SET_TOOL_VOLTAGE = 2,  ///< Set tool voltage
+};
```
fmauch, line 48: Is there a certain intention why this enum is not a private class member? It seems like this is only used internally and not necessarily meant to be used externally, since you wrapped the different commands into individual functions.

fmauch (commented):
In general seems very clean. Tested together with https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/567 and works fine.

Not approving because of a couple of discussion points left.

src/ur/ur_driver.cpp
fmauch: I took the liberty of implementing this, as I wanted to have this for debugging while testing, anyway. See e86b1f8857690de81f06fda0665941ee3b20f7fe

src/ur/ur_driver.cpp
fmauch: I took the liberty of implementing this, as I wanted to have this for debugging while testing, anyway. See e86b1f8857690de81f06fda0665941ee3b20f7fe

src/ur/ur_driver.cpp
fmauch: I took the liberty of implementing this, as I wanted to have this for debugging while testing, anyway. See e86b1f8857690de81f06fda0665941ee3b20f7fe

src/ur/ur_driver.cpp
urmahp: I agree that it would be better to print an error message and return false.

include/ur_client_library/control/script_command_interface.h
urmahp: No, I think I just copied the setup from one of the other classes. Since it is only used inside the class it definitely makes sense to make it private.

fmauch (approved):
LGTM, Thanks @urmahp

Issue #113, opened 2022-08-29
RTDE protocol v1 is not supported in setup outputs


urmahp:
When trying to setup outputs in the rtde client using protocol v1 the rtde client will crash and the output when it is crashing can be seen below.   

```bash
./build/examples/rtde_client_example 
WARN /src/Universal_Robots_Client_Library/include/ur_client_library/comm/pipeline.h 406: No realtime capabilities found. Consider using a realtime system for better performance 
terminate called after throwing an instance of 'urcl::UrException'
  what():  Could not parse received package. This can occur if the driver is started while the robot is booting - please restart the driver once the robot has finished booting. If the problem persists after the robot has booted, please contact the package maintainer.
Aborted (core dumped)
****
```

This happens because the control_package_setup_outputs class tries to parse a buffer that is longer than the received one, because the answer from the robot is different whether you use protocol v1 or v2. The error occurs in the [setupOutputs function](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/src/rtde/rtde_client.cpp#L237-L310), when trying to parse the package received from the robot. I will update the client library to support protocol v1 as well, by updating the control_package_setup_outputs class to handle parsing protocol v1.

fmauch:
Should be resolved by #114 -> closing

PR #114, opened 2022-08-29
Fixed parsing of incomming packages when using rtde protocol v1

urmahp:
The received rtde packages should be parsed slightly different whether we use protocol v1 or v2.

This will fix issue #113

### DIFF
--- a/src/rtde/control_package_setup_outputs.cpp
+++ b/src/rtde/control_package_setup_outputs.cpp
@@ -34,16 +34,30 @@ namespace rtde_interface
 {
 bool ControlPackageSetupOutputs::parseWith(comm::BinParser& bp)
 {
-  bp.parse(output_recipe_id_);
-  bp.parseRemainder(variable_types_);
+  if (protocol_version_ == 2)
+  {
+    bp.parse(output_recipe_id_);
+    bp.parseRemainder(variable_types_);
+  }
+  else if (protocol_version_ == 1)
+  {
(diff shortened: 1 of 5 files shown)


src/rtde/control_package_setup_outputs.cpp
```diff
@@ -34,16 +34,30 @@ namespace rtde_interface
...
-  bp.parseRemainder(variable_types_);
+  if (protocol_version_ == 2)
+  {
+    bp.parse(output_recipe_id_);
+    bp.parseRemainder(variable_types_);
+  }
+  else if (protocol_version_ == 1)
+  {
+    bp.parseRemainder(variable_types_);
+  }
```
fmauch, line 45: Maybe adding a plain `else` tree here if UR ever decides to to a protocol version 3 would be good. This way it will be obvious why, when things go south.

fmauch (commented):
Looks good to me! As mentioned inside the code comments, it might be a good point to also add some failsafes for the case when there's more than version 1 and 2. @urmahp would you agree?

src/rtde/control_package_setup_outputs.cpp
urmahp: That is a good idea, I have added a plain else statement.

fmauch:
> Should I squash the commits to one commit?

We do that when merging. We usually use the "squash and merge" strategy.

fmauch (approved):
LGTM if pipeline succeeds.

PR #117, opened 2022-09-23
Update atomicops.h

urrsk:
Fix the url in the comment regarding POSIX semaphores to fix error in the CI


fmauch:
That would be an alternate (better) way of what I wanted to suppress in #108. Thanks!

fmauch (approved):
With https this url actually exists -> fine

Issue #115, opened 2022-09-15
Dashboard Client seg faults when switching from remote mode to local mode


hugols16:
On the e-series, the Teach Pendant must be in remote mode to control the arm. Detecting whether it's in remote mode or not can be done via the dashboard client. The problem resides when using this to monitor whether it's in remote mode, and expecting the end user to switch to local modes at times (for jogging, freedrive, etc.). When the end user switches to local mode on the Teach Pendant, the dashboard server closes the connection, which is not handled properly by the Dashboard Client, and it ultimately segfaults. This can be tested very simply:

1. Start your program in remote mode
2. Periodically check if the Teach Pendant is in remote mode
3. Manually switch to local mode
4. It segfaults

fmauch:
> On the e-series, the Teach Pendant must be in remote mode to control the arm.

This is not really true.

However, we know that changing between local and remote mode does cause problems with the dashboard client, but I was not aware that this actually causes segfaults. I can currently not reproduce this, as the disconnect does not happen with URSim, but I'll give that a bug label nevertheless and try to reproduce with a real robot later.

fmauch:
I could not reproduce a segfault. When I start the driver while being in remote_control mode, then switch to local mode and try to send something to the dashboard server, I get

```
[dashboard_client-2] [ERROR] [1663768349.973294092] [UR_Client_Library]: Attempt to write on a non-connected socket
[dashboard_client-2] [ERROR] [1663768349.973329027] [Dashboard_Client]: Service Call failed: 'Failed to send request to dashboard server. Are you connected to the Dashboard Server?'
```
Then calling
```
ros2 service call /dashboard_client/connect std_srvs/srv/Trigger
```
brings me back to a working dashboard client.

I'm not saying that the issue does not exist, but I can't currently reproduce it. Could you tell us a bit more about your setup? Which ROS version, source build or binary packages, Polyscope version....

urmarp:
This issue has been addressed in PR #118

fmauch:
@urmarp so you managed to reproduce the issue? Could you tell us a bit more about that, please?

urmarp:
Yes, I could reproduce the error on a real robot with SW 5.12 by following the steps @hugols16 mentioned. The problem was that the receive timeout was not set in the dashboard client.

fmauch:
[...]

PR #118, opened 2022-09-26
Initialized receive timeout and changed exception to warning

urmarp:
This will fix issue #115

### DIFF
--- a/src/ur/dashboard_client.cpp
+++ b/src/ur/dashboard_client.cpp
@@ -50,6 +50,12 @@ bool DashboardClient::connect()
     URCL_LOG_INFO("%s", read().c_str());
     ret_val = true;
   }
+
+  timeval tv;
+  tv.tv_sec = 1;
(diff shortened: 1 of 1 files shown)


fmauch (commented):
It definitively makes sense to initialize the timeout by default.

Concerning the exception vs. warning output: It probably is a matter of control flow definition. The aim was that if the read fails, we clearly communicate that to the caller. By simply returning an empty string the caller has to explicitly check whether the returned string does make sense. A simple check against an empty string might be misleading, as this would require that a dashboard call would never return an empty string (which might be true, actually). By throwing an exception a caller can make a distinct decision what to do in that case by defining an exit strategy inside a try/catch block.

Just for the sake of clarity: I'm not saying that this is the one and only way to handle this, but this is a workflow we've used at multiple parts of this library, so far.

fmauch (approved):
LGTM, thanks @urmarp

PR #125, opened 2022-10-24
Added tests for the rtde interface clasess

urmahp:
This adds tests for
* control package pause
* control package start
* control package setup outputs
* control package setup inputs
* get urcontrol version
* request protocol version
* rtde writer

Added more tests to
* test_rtde_data_package
* test_rtde_parser
* test_rtde_client

Added functionality to rtde_client
* get_target_frequency function and warning when it is not possible to set target frequency
* Read timeout in isRobotBooted is now based on target frequency
* Added a wait on 5 seconds to receive answer from pause request. This will prevent the driver from crashing once in a while when pausing the rtde client

Added functionality for rtde_writer
* Added checks for values and pins before sending to the robot
* Removed unnecessary sleep in desctructor

Added pipeline changes, so that the test output is shown in the pipeline in case of failure

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
       - name: build
         run: cmake --build build --config Debug
       - name: test
-        run: cd build && make test
+        run: cd build && ctest --output-on-failure
       - name: install gcovr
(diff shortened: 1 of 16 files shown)


fmauch (approved):
Thanks for implementing this!

PR #126, opened 2022-10-24
dashboard_client: Allow a larger timeout on initial connection

fmauch:
Apparently, if we connect to the socket right after it became available, it can happen, that we don't get a "connected" message from the dashboard server in a second.

I tested this with a retry timeout of 1ms [here](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/1d9b9750128b31081f57ecf466f8acc7e7538f60/src/comm/tcp_socket.cpp#L106). Then, when launching the dashboard_client before starting URSim made the client trying to connect each millisecond. I read the `Connected: Universal Robots Dashboard Server` message roughly 1.5 seconds after the last `Failed to connect to robot on IP...` output. This seems to be the minimum time we need to wait for the bootup message after the socket has been opened (at least on my machine with URSim).

Therefore, I set the timeout for the initial `read` call to 10 seconds and added a retry mechanism to that. I don't think this should happen which is why I added a warning output.

### DIFF
--- a/src/ur/dashboard_client.cpp
+++ b/src/ur/dashboard_client.cpp
@@ -45,13 +45,31 @@ bool DashboardClient::connect()
     return false;
   }
   bool ret_val = false;
-  if (TCPSocket::setup(host_, port_))
-  {
-    URCL_LOG_INFO("%s", read().c_str());
-    ret_val = true;
-  }
 
   timeval tv;
+
+  while (not ret_val)
+  {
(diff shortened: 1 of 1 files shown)


urmahp (approved):
LGTM. 

I have verified that it works with the increased timeout and that it doesn't work with the original timeout of 1 seconds when the robot simulator has just finished booting.

PR #127, opened 2022-10-29
Dashboard commands, Docker Image and CI step for running the examples

urrsk:
This PR started with adding a example for getting the calibration info out from the primary interface. Then I want it to run in the CI for continue test that it could run.
After I added a step running the examples, I realised that the other examples need some adjustments, including some calls through the dashboard server, to turn the Simulated robot into run mode.
Therefore includes this PR also functionalities to run Dashboard command for this purpose, just like it is made in the ROS drivers.
In addition, the CI also was update to use the docker imagel


examples/dashboard_example.cpp
```diff
@@ -0,0 +1,127 @@
...
+  {
+    URCL_LOG_ERROR("Could not send BreakeRelease command");
+    return 1;
+  }
+
+  // Load existing program
+  const std::string program_file_name_to_be_loaded("wait_program.urp");
+  if (!my_dashboard->commandLoadProgram(program_file_name_to_be_loaded))
+  {
+    URCL_LOG_ERROR("Could not load %s program", program_file_name_to_be_loaded.c_str());
```
urmahp, line 85:
```suggestion
    URCL_LOG_ERROR("Could not load %s program. The wait_program.urp can be found inside the examples/resources/ folder", program_file_name_to_be_loaded.c_str());
```
I think we should add the program to the examples/resources/ folder and the point towards the folder if it fails to load the program.

include/ur_client_library/ur/dashboard_client.h
```diff
@@ -78,6 +78,131 @@ class DashboardClient : public comm::TCPSocket
...
+   * \return True succeeded
+   */
+  bool commandPowerOn(unsigned int timeout = 1200);
+
+  /*!
+   * \brief Send Brake release command
+   *
+   * \return True succeeded
+   */
+  bool commandBreakeRelease();
```
urmahp, line 139:
```suggestion
  bool commandBrakeRelease();
```
Make the function name the same as the dashboard command

src/ur/dashboard_client.cpp
```diff
@@ -113,4 +119,124 @@ std::string DashboardClient::sendAndReceive(const std::string& text)
...
+{
+  return sendRequest("power off", "Powering off") && waitForReply("robotmode", "Robotmode: POWER_OFF");
+}
+
+bool DashboardClient::commandPowerOn(unsigned int timeout)
+{
+  return retryCommand("power on", "Powering on", "robotmode", "Robotmode: IDLE", timeout);
+}
+
+bool DashboardClient::commandBreakeRelease()
```
urmahp, line 191:
```suggestion
bool DashboardClient::commandBrakeRelease()
```
Reflect header changes.

examples/dashboard_example.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  // Power it on
+  if (!my_dashboard->commandPowerOn())
+  {
+    URCL_LOG_ERROR("Could not send Power on command");
+    return 1;
+  }
+
+  // Release the brakes
+  if (!my_dashboard->commandBreakeRelease())
```
urmahp, line 75:
```suggestion
  if (!my_dashboard->commandBrakeRelease())
```
Reflect changes to dashboard client.

examples/full_driver.cpp
```diff
@@ -56,15 +58,71 @@ void handleRobotProgramState(bool program_running)
...
+
+  // Power it on
+  if (!my_dashboard->commandPowerOn())
+  {
+    URCL_LOG_ERROR("Could not send Power on command");
+    return 1;
+  }
+
+  // Release the brakes
+  if (!my_dashboard->commandBreakeRelease())
```
urmahp, line 101:
```suggestion
  if (!my_dashboard->commandBrakeRelease())
```
Reflect changes to dashboard client.

urmahp (commented):
I would add the wait_program.urp to the resources folder inside the examples, since it is required by the dashboard_example.cpp and also add to the error message if it fails to load the program, that the program can be found in the resources folder. 

I think that we should add functions for all or at least most of the dashboard commands available, but this doesn't necessarily have to be in this PR. 

Other than above comments it looks very good to me and I have tested that all the examples are working properly, with a simulated robot.

examples/dashboard_example.cpp
urrsk: Good points. I like the error message as well.
But the urp program is different between cb3 and e-series so I chose to place it in the folder directly.
An alternative could be to copy it into the program folder just before the test.

urrsk: Though, lets start with this.
The purpose of the example is to show how to do it and the program only consist of a wait node

examples/dashboard_example.cpp
```diff
@@ -0,0 +1,127 @@
...
+  {
+    URCL_LOG_ERROR("Could not send BreakeRelease command");
```
urrsk, line 77:
```suggestion
    URCL_LOG_ERROR("Could not send BrakeRelease command");
```

examples/full_driver.cpp
```diff
@@ -56,15 +58,71 @@ void handleRobotProgramState(bool program_running)
...
+  {
+    URCL_LOG_ERROR("Could not send BreakeRelease command");
```
urrsk, line 103:
```suggestion
    URCL_LOG_ERROR("Could not send BrakeRelease command");
```

src/ur/dashboard_client.cpp
```diff
@@ -113,4 +119,124 @@ std::string DashboardClient::sendAndReceive(const std::string& text)
...
+}
+
+bool DashboardClient::commandPowerOff()
+{
+  return sendRequest("power off", "Powering off") && waitForReply("robotmode", "Robotmode: POWER_OFF");
+}
+
+bool DashboardClient::commandPowerOn(unsigned int timeout)
+{
+  return retryCommand("power on", "Powering on", "robotmode", "Robotmode: IDLE", timeout);
```
urrsk, line 188: @fmauch This was the retry function I talked about

include/ur_client_library/ur/dashboard_client.h
```diff
@@ -78,6 +78,131 @@ class DashboardClient : public comm::TCPSocket
...
+   * \brief Send Power off command
+   *
+   * \return True succeeded
+   */
+  bool commandPowerOff();
+
+  /*!
+   * \brief Send Power on command
+   *
+   * \param timeout Timeout in seconds
```
fmauch, line 128: Maybe a comment on the header, why this is necessary would be great.

fmauch (changes requested):
Thanks @urrsk for pushing this from the drivers to the library. I think, it is worth modernizing a couple of things while we're at it, see my comments. As discussed, I'll add some suggestions to the PR directly.

fmauch:
@urrsk I've included most of your comments from https://github.com/urrsk/Universal_Robots_Client_Library/pull/1 and I think I would be happy with the state this is in. I tend to disagree with https://github.com/urrsk/Universal_Robots_Client_Library/pull/1#discussion_r1028017404

This feels like hitting the "using exceptions as control flow is an anti-pattern" too hard. That might be a more general question though, how we would like to structure this library.

fmauch (approved):
Thank you @urrsk and @urmahp for the great work. This should greatly robustify things around the dashboard server.

I'll clean up history a bit and will then merge it.

fmauch:
Merging despite failing pipeline on humble, because of #132.

PR #129, opened 2022-11-07
Added tests for the comm classes

urmahp:
This add tests for
* bin parser
* package serializer
* TCP socket
* stream
* producer
* pipeline

Fixed check size in bin parser, so that it works with starndard data types

Added recconection time variable to TCP socket, this will control the time in between connection attempts if the server is unavailable.


urmahp:
I found the issue. Since the consumer is running in a thread sometimes the consumer wasn't started when the message was send to the client and therefore the test failed, this has been fixed with a retry mechanism, so the message is resend to the client if the first one wasn't consumed.

fmauch (approved):
After fixing the segfault in the pipeline test this seems fine. Thank you @urmahp

PR #130, opened 2022-11-08
Catch Exception by reference and not by value

fmauch:
In general this is a good idea and it will silence a compiler warning.


urmahp (approved):
LGTM.

PR #131, opened 2022-11-10
Add a colcon.pkg file for proper ament integration

fmauch:
This will help making the package available in the ament build system. This will effecticely advertise the hooks that we install in our `CMakeLists.txt`

This will effectively enable https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/316 to work properly.


urmahp (approved):
LGTM and I have tested that the ros2 driver is able to locate the control script from the client library with these changes.

PR #134, opened 2022-11-28
Add a prerelease check that calls `bloom-generate`

fmauch:
Since I had problems in the `bloom-generate` step when releasing 1.3.0 and similar things have happened in the past, I think it is a good idea to add a check to new PRs to prevent such things in the future.

This PR also contains a fix for the regression currently living on the master branch introduced in #127.


RobertWilbrandt (approved):
This seems like a reasonable approach

PR #136, opened 2022-12-28
Changed num_retries from static to an unsigned int

urmahp:
The num_retries wouldn't be reset within each function, but it will now so the correct number of retries is performed in each function, before throwing an exception.


fmauch (approved):
Looks good to me. Thank you @urmahp

PR #137, opened 2022-12-28
Improve primary client

urmarp:
Improved primary interface, this includes following changes:

 * Added message handlers for messages received through primary client

 * Changed primary pipeline and full driver example

 * Added datatype: vector6f_t and vector6uin8_t

 * Added parsing of vector6f_t

 * Changed timeout in producer

 * Added parsing of more messages from primary interface

 * Added calibration check in constructor of primary client

 * Added calibration checksum to urdriver constructor

 * Added test for primary client and primary parser

 * Added funtionality to set robot as simulated or real

 * Added wait in full driver example


fmauch:
@urmarp this would be easier to review if these were multiple commits.

include/ur_client_library/comm/producer.h
```diff
@@ -105,24 +108,24 @@ class URProducer : public IProducer<T>
...
         // reset sleep amount
-        timeout_ = std::chrono::seconds(1);
+        timeout_ = std::chrono::milliseconds(100);
         BinParser bp(buf, read);
         return parser_.parse(bp, products);
       }
 
-      if (!running_)
-        return true;
-
```
fmauch, line 121: Why exactly did you drop this? I don't remember out of my head why we have it there, but I would be hesitant deleting this without further information.

fmauch (changes requested):
Before reviewing this any further: Could you please elaborate what you have done? What is the handler concept and why did you add this?

Again, splitting this into multiple commits (maybe with commit messages going into more detail) might make things a lot clearer.

urmarp:
> Before reviewing this any further: Could you please elaborate what you have done? What is the handler concept and why did you add this?
> 
> Again, splitting this into multiple commits (maybe with commit messages going into more detail) might make things a lot clearer.

I have split the commit into multiple commits and tried to explain the thoughts behind the changes. 

Added message handlers for the messages from the primary interface. These handlers save the data from the latest message of each type, which can then be accessed through the primary client. The handlers also enable the possibility to handle the data differently for each message. The primary client can through its consumer get the message handlers and from the message handlers get the latest message of each type.

fmauch:
Closing this in favor of #186

PR #139, opened 2023-01-30
Specifically set RTDE pipeline producer to FIFO scheduling

fmauch:
Regardless of the actual kernel used, setting the producer thread to FIFO scheduling seems to improve robustness.

### DIFF
--- a/include/ur_client_library/comm/pipeline.h
+++ b/include/ur_client_library/comm/pipeline.h
@@ -376,7 +376,6 @@ private:
       const int max_thread_priority = sched_get_priority_max(SCHED_FIFO);
       setFiFoScheduling(this_thread, max_thread_priority);
     }
-    std::ifstream realtime_file("/sys/kernel/realtime", std::ios::in);
     std::vector<std::unique_ptr<T>> products;
     while (running_)
     {
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -80,6 +80,7 @@ bool setFiFoScheduling(pthread_t& thread, const int priority)
     URCL_LOG_INFO("SCHED_FIFO OK, priority %i", params.sched_priority);
     if (params.sched_priority != priority)
     {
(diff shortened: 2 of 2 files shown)


include/ur_client_library/comm/pipeline.h
```diff
@@ -349,62 +365,18 @@ class Pipeline
...
-      {
-        URCL_LOG_ERROR("Could not get maximum thread priority for producer thread");
-      }
-    }
-    else
-    {
-      URCL_LOG_WARN("No realtime capabilities found. Consider using a realtime system for better performance");
+      setFiFoScheduling(this_thread, max_thread_priority);
     }
+    std::ifstream realtime_file("/sys/kernel/realtime", std::ios::in);
```
RobertWilbrandt, line 379: Does this do anything?

src/helpers.cpp
```diff
@@ -0,0 +1,88 @@
...
+  {
+    URCL_LOG_ERROR("Scheduling is NOT SCHED_FIFO!");
+    return false;
+  }
+  else
+  {
+    URCL_LOG_INFO("SCHED_FIFO OK, priority %i", params.sched_priority);
+    if (params.sched_priority != priority)
+    {
+      return false;
```
RobertWilbrandt, line 83: I would expect there to be a `URCL_LOG_ERROR` here too, otherwise it will just fail without telling the user why.

include/ur_client_library/comm/pipeline.h
fmauch: No, this is a copy_paste error from copying the check where we actually checked whether the system does run a realtime kernel. Since we want to use FIFO scheduling independent of this, this line can be removed.

src/helpers.cpp
fmauch: Done in latest commit.

include/ur_client_library/comm/pipeline.h
fmauch: Removed that line in the latest commit.

RobertWilbrandt (approved):

PR #146, opened 2023-04-27
Add a cmake option to activate address sanitizers

fmauch:
In an effort to track down memory issues, I started activating address sanitizers on the urcl. I think, we should activate them by default inside CI in order to spot issues early on.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE RelWithDebInfo)
 endif()
 
+option(WITH_ASAN "Compile with address sanitizer support" OFF)
 
 ##
 ## Check C++11 support / enable global pedantic and Wall
@@ -51,6 +52,10 @@ add_library(urcl SHARED
 add_library(ur_client_library::urcl ALIAS urcl)
 target_compile_options(urcl PRIVATE -Wall -Wextra -Wno-unused-parameter)
 target_compile_options(urcl PUBLIC ${CXX17_FLAG})
+if(WITH_ASAN)
+  target_compile_options(urcl PUBLIC -fsanitize=address)
+  target_link_options(urcl PUBLIC -fsanitize=address)
+endif()
 target_include_directories( urcl PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include>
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -8,6 +8,12 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE RelWithDebInfo)
 endif()
 
+option(WITH_ASAN "Compile with address sanitizer support" OFF)
+if(WITH_ASAN)
+  add_compile_options(-fsanitize=address)
+  add_link_options(-fsanitize=address)
+endif()
```
RobertWilbrandt, line 15: I am not a big fan of setting these flags globally, especially for libraries. I think
```
target_compile_options(urcl PRIVATE -fsanitize=address)
target_link_options(urcl PRIVATE -fsanitize=address)
```
would be the better choice.

fmauch: In principle I would agree. However, using this inside the main CMakeLists.txt
   ```cmake
   ...
   add_library(ur_client_library::urcl ALIAS urcl)
   if(WITH_ASAN)
     target_compile_options(urcl PRIVATE -fsanitize=address)
     target_link_options(urcl PRIVATE -fsanitize=address)
   endif()
   target_compile_options(urcl PRIVATE -Wall -Wextra -Wno-unused-parameter)
   target_compile_options(urcl PUBLIC ${CXX17_FLAG})
   ...
   ```
leads to this (without further modifications)
   ```
   build/examples/dashboard_example 192.168.56.101
   ==416292==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
   ```
So, one has to add the targeted sanitizer to **each** target, as the targets depending on urcl won't work (won't run) otherwise. Since from a maintenance perspective it seems not right to copy those four lines to each of our many targets I would suggest sticking with the global flag. It should only be activated manually for testing purposes, so I guess, this would be valid price to pay.

RobertWilbrandt: Can you try using
```
  target_compile_options(urcl PUBLIC -fsanitize=address)
  target_link_options(urcl PUBLIC -fsanitize=address)
```
instead? From my understanding that should serve exactly this role, e.g. propagating to all targets linking to `urcl`.

fmauch: Sounds like a sane idea, will try that.

fmauch:
@RobertWilbrandt I've updated the PR using targeted compile options. This should now probably be fine for merging

RobertWilbrandt (approved):

PR #149, opened 2023-06-30
Make URSim log files available as artifacts

fmauch:
This is an attempt to make debugging failing tests easier.


fmauch:
@urrsk I didn't use the dashboard_client functionality as it was easier to do it using a direct `netcat` call from the command line, but I think it is still fine to keep this inside this PR. We'll simply merge this using both commits.

urrsk (approved):

urrsk:
Maybe the script should be placed in the script folder..

PR #150, opened 2023-06-30
Remove Foxy from CI

fmauch:
Since the Foxy prerelease tests start throwing errors and Foxy is EOL, I think we can drop Foxy from our CI.


urrsk (approved):
Agree

PR #103, opened 2022-06-09
Implemented spline interpolation in joint space

urmahp:

PR #151, opened 2023-06-30
Spline merge

urrsk:
Added Forward Spline interpolation to make the robot handle spline by it self 
An attempt to merge #103

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,13 @@ jobs:
             BEFORE_INIT: 'apt-get update -qq && apt-get install -y iproute2 iputils-ping && ip addr && ping -c5 192.168.56.101'
             CTEST_OUTPUT_ON_FAILURE: 1
             ROBOT_MODEL: 'ur5'
-            URSIM_VERSION: '3.12.1'
+            URSIM_VERSION: '3.14.3'
             PROGRAM_FOLDER: 'tests/resources/dockerursim/programs/cb3'
           - DOCKER_RUN_OPTS: --network ursim_net
             BEFORE_INIT: 'apt-get update -qq && apt-get install -y iproute2 iputils-ping && ip addr && ping -c5 192.168.56.101'
             CTEST_OUTPUT_ON_FAILURE: 1
             ROBOT_MODEL: 'ur5e'
-            URSIM_VERSION: '5.5.1'
+            URSIM_VERSION: '5.9.4'
             PROGRAM_FOLDER: 'tests/resources/dockerursim/programs/e-series'
 
     steps:
--- a/examples/spline_example.cpp
+++ b/examples/spline_example.cpp
@@ -227,34 +227,6 @@ int main(int argc, char* argv[])
     time.push_back(s_time[i]);
   }
 
-  // QUADRATIC
-  SendTrajectory(p, std::vector<vector6d_t>(), std::vector<vector6d_t>(), time, true);
-
-  g_trajectory_running = true;
-  while (g_trajectory_running)
-  {
-    std::unique_ptr<rtde_interface::DataPackage> data_pkg = g_my_driver->getDataPackage();
-    if (data_pkg)
-    {
-      // Read current joint positions from robot data
-      if (!data_pkg->getData("actual_q", g_joint_positions))
-      {
-        // This throwing should never happen unless misconfigured
-        std::string error_msg = "Did not find 'actual_q' in data sent from robot. This should not happen!";
-        throw std::runtime_error(error_msg);
-      }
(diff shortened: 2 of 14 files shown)


examples/tool_contact_example.cpp
```diff
@@ -114,6 +114,8 @@ int main(int argc, char* argv[])
   const bool HEADLESS = true;
   g_my_driver.reset(new UrDriver(robot_ip, SCRIPT_FILE, OUTPUT_RECIPE, INPUT_RECIPE, &handleRobotProgramState, HEADLESS,
                                  std::move(tool_comm_setup), CALIBRATION_CHECKSUM));
+  g_my_driver->setKeepaliveCount(5);  // decrease realtime requirements since example is running in
+                                      // CI
```
fmauch, line 118: Maybe we should use a more explicit comment:

```suggestion
  g_my_driver->setKeepaliveCount(5);  // This is for example purposes only. This will make the example running more 
                                      // reliable on non-realtime systems. Do not use this in productive applications.
```

examples/spline_example.cpp
```diff
@@ -0,0 +1,287 @@
...
+const std::string INPUT_RECIPE = "examples/resources/rtde_input_recipe.txt";
+const std::string CALIBRATION_CHECKSUM = "calib_12788084448423163542";
+
+std::unique_ptr<UrDriver> g_my_driver;
+std::unique_ptr<DashboardClient> g_my_dashboard;
+vector6d_t g_joint_positions;
+
+void SendTrajectory(const std::vector<vector6d_t>& p_p, const std::vector<vector6d_t>& p_v,
+                    const std::vector<vector6d_t>& p_a, const std::vector<double>& time, bool use_spline_interpolation_)
+{
```
fmauch, line 44: There could be an assertion here that p_p has the same size as time, as this is a hard requirement.

examples/spline_example.cpp
```diff
@@ -0,0 +1,287 @@
...
+    URCL_LOG_ERROR("Could not send Power off command");
+    return 1;
+  }
+
+  // Power it on
+  if (!g_my_dashboard->commandPowerOn())
+  {
+    URCL_LOG_ERROR("Could not send Power on command");
+    return 1;
+  }
```
fmauch, line 145: What exactly is the motivation behind this? When doing tests with a real robot that seems a bit weird that the robot switches off after starting that program. Also, it takes a lot of time to boot the robot again (which seems unnecessary). If we reduce this to only switching the robot an and releasing the brakes, it would have the same effect that we know the robot will be running afterwards without the additional switch off / switch on thing.

src/control/trajectory_point_interface.cpp
```diff
@@ -85,6 +90,86 @@ bool TrajectoryPointInterface::writeTrajectoryPoint(const vector6d_t* positions,
...
 
+bool TrajectoryPointInterface::writeTrajectorySplinePoint(const vector6d_t* positions, const vector6d_t* velocities,
+                                                          const vector6d_t* accelerations, const float goal_time)
+{
+  if (client_fd_ == -1)
+  {
+    return false;
+  }
+
+  control::TrajectorySplineType spline_type = control::TrajectorySplineType::SPLINE_QUINTIC;
```
fmauch, line 101: I think this can be cleaned up a bit in such a way that we initialize the `spline_type` as `SPLINE_QUADRATIC` and change it to a higher order in the if clauses when velocity / acceleration information is given.

fmauch (commented):
Thank you @urrsk for cleaning this up. There are a couple of comments left I've done in #103. I've added them to this PR.

fmauch:
Note to myself: @urmahp should be noted as coauthor in the merge.

.github/workflows/ci.yml
```diff
@@ -12,13 +12,13 @@ jobs:
             BEFORE_INIT: 'apt-get update -qq && apt-get install -y iproute2 iputils-ping && ip addr && ping -c5 192.168.56.101'
             CTEST_OUTPUT_ON_FAILURE: 1
             ROBOT_MODEL: 'ur5'
-            URSIM_VERSION: '3.12.1'
+            URSIM_VERSION: '3.14.3'
```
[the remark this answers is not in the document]
urrsk: Yes, Matrix and scalar product was added at the same time

examples/spline_example.cpp
urrsk: I copied it from the full_driver_example. 
Think it was to ensure that same starting point. But agree it could be optimized

fmauch (approved):
This is looking good and I tested it with the example and the ROS1 driver PR.

PR #152, opened 2023-07-01
Add codecov.yml to exclude test and examples folders

urrsk:
Avoid punishing commits that includes examples and unit test


fmauch (approved):
Seems fair

PR #153, opened 2023-07-01
Make URSim log files available as artifacts also for the CI-industrial

urrsk:


fmauch (approved):
Yes, that makes sense, thank you!

PR #155, opened 2023-07-05
Install start ursim

fmauch:
This PR installs the start_ursim script to the package's install space. This way, it can be reused in downstream packages such as the ROS drivers.


fmauch:
This seems to work with https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/640 and https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/733

RobertWilbrandt (approved):
Seems like a great step to remove these redundant files!

Issue #144, opened 2023-03-16
RTDE client fail if timestamp is not present in output recipe


jornb:
Error is `Package of type 85 was not parsed completely!`, because the robot sends timestamp which is not read.

In `readRecipe`, "timestamp" should be added if it isn't there already.

Introduced in #98

urmahp:
Thank you for reporting this. I have been able to reproduce the issue and will look at implementing a fix today.

urmahp:
Thank you very much @jornb for reporting this, a fix has been implemented in #168.

PR #168, opened 2023-08-23
Timestamp is now added to the output recipe correctly

urmahp:
Added test to verify the addition of timestamp

Only disconnect if client is initialized

This will fix the issue in #144


include/ur_client_library/rtde/rtde_client.h
```diff
@@ -208,7 +208,7 @@ class RTDEClient
   constexpr static const double CB3_MAX_FREQUENCY = 125.0;
   constexpr static const double URE_MAX_FREQUENCY = 500.0;
 
-  std::vector<std::string> readRecipe(const std::string& recipe_file);
+  std::vector<std::string> readRecipe(const std::string& recipe_file, bool output_recipe = false);
```
urrsk, line 211: Suggest we a using a an enum so it is more readable to that the boolean is doing.
And comments won't hurt either

src/rtde/rtde_client.cpp
```diff
@@ -559,11 +555,29 @@ std::vector<std::string> RTDEClient::readRecipe(const std::string& recipe_file)
...
+    throw UrException(msg.str());
+  }
+
   std::string line;
   while (std::getline(file, line))
   {
     recipe.push_back(line);
   }
+
+  // Add timestamp to rtde output recipe, used to check if robot is booted
```
urrsk, line 573:
```suggestion
  // Add timestamp to rtde output recipe, if not already existing.
  // The timestamp is used to check if robot is booted or not
```

urrsk (changes requested):

src/rtde/rtde_client.cpp
```diff
@@ -37,8 +37,8 @@ namespace rtde_interface
 RTDEClient::RTDEClient(std::string robot_ip, comm::INotifier& notifier, const std::string& output_recipe_file,
                        const std::string& input_recipe_file, double target_frequency)
   : stream_(robot_ip, UR_RTDE_PORT)
-  , output_recipe_(readRecipe(output_recipe_file))
-  , input_recipe_(readRecipe(input_recipe_file))
+  , output_recipe_(readRecipe(output_recipe_file, RecipeType::OUTPUT_RECIPE))
```
urrsk, line 40:
```suggestion
  , output_recipe_(ensureTimestampIsPresent(readRecipe(output_recipe_file)))
```

- [ ] remove enum as well

urrsk (approved):

PR #170, opened 2023-08-24
Only warn if system is not setup for FIFO scheduling

RobertWilbrandt:
Based on https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/654

I agree that an error message seems too hard here, as you are still able to operate the driver even if you are not setup for FIFO scheduling. We also shouldn't error out twice, if we cannot change the scheduling parameters the second check seems redundant.


fmauch (approved):
From my memory we explicitly decided to make this an error as we wanted to emphasis that we expect things to go wrong when not using a PREEMPT_RT kernel. However, also from the experience I've collected over the years developing with a standard kernel without any special setup I agree that going to a warning seems fine.

PR #141, opened 2023-02-08
Add support for setting socket max num tries and reconnect timeout 

michaelaeriksen:
Fixes #140

PR #172, opened 2023-08-29
Add support for setting socket max num tries and reconnect timeout

### LINKED ISSUE #140: TCPSocket loops forever - add timeout or max num tries
In TCPSocket::setup we are looping forever until connected. It would be great to add a timeout or max_num_tries as an argument.

fmauch:
This continues #141 and fixes #140.

## ToDo:
- [ ] Add tests
  - [ ] Attempt to connect to a non-running robot with specified timeout and max_num_retries


fmauch:
Going back to draft because of failing tests

include/ur_client_library/comm/tcp_socket.h
```diff
@@ -59,7 +59,7 @@ class TCPSocket
   }
   virtual void setOptions(int socket_fd);
 
-  bool setup(std::string& host, int port);
+  bool setup(std::string& host, int port, size_t max_num_tries = 0);
```
urmahp, line 62: I know it is not related to this PR, but the host is parsed as a non const reference even though it is not modified within the function. I think it would make sense to add the const keyword to host and all the other parameters as they are not modified within the function. 
```suggestion
  bool setup(const std::string& host, const int port, const size_t max_num_tries = 0);
```

include/ur_client_library/comm/pipeline.h
```diff
@@ -287,9 +292,16 @@ class Pipeline
...
 
-  void init()
+  /*!
+   * \brief Initialize the pipeline. Internally calls setup of producer and consumer.
+   *
+   * \param max_num_tries Maximum number of connection attempts before counting the connection as
+   * failed. Unlimited number of attempts when set to 0.
+   * \param reconnection_time time in between connection attempts to the server
+   */
+  void init(size_t max_num_tries = 0, std::chrono::milliseconds reconnection_time = std::chrono::seconds(10))
```
urmahp, line 302: Add const keyword

```suggestion
  void init(const size_t max_num_tries = 0, const std::chrono::milliseconds reconnection_time = std::chrono::seconds(10))
```

include/ur_client_library/rtde/rtde_client.h
```diff
@@ -228,7 +233,8 @@ class RTDEClient
   // the robot is booted.
   std::vector<std::string> ensureTimestampIsPresent(const std::vector<std::string>& output_recipe) const;
 
-  void setupCommunication();
+  void setupCommunication(size_t max_num_tries = 0,
+                          std::chrono::milliseconds reconnection_time = std::chrono::seconds(10));
```
urmahp, line 237: Add const keyword
```suggestion
                        const std::chrono::milliseconds reconnection_time = std::chrono::seconds(10));
```

include/ur_client_library/rtde/rtde_client.h
```diff
@@ -228,7 +233,8 @@ class RTDEClient
   // the robot is booted.
   std::vector<std::string> ensureTimestampIsPresent(const std::vector<std::string>& output_recipe) const;
 
-  void setupCommunication();
+  void setupCommunication(size_t max_num_tries = 0,
```
urmahp, line 236: Add const keyword.
```suggestion
  void setupCommunication(const size_t max_num_tries = 0,
```

include/ur_client_library/comm/tcp_socket.h
```diff
@@ -50,7 +50,7 @@ class TCPSocket
 private:
   std::atomic<int> socket_fd_;
   std::atomic<SocketState> state_;
-  std::chrono::seconds reconnection_time_;
+  std::chrono::milliseconds reconnection_time_;
```
urmahp, line 53: I know that reconnection_time was not added in this PR, but I suggest that we add the reconnection_time as an input argument to the setup function instead of a member variable. Then max_num_tries and reconnection_time can be modified in the same way and both variables are only used within the setup function.

urmahp (commented):
Overall it looks very good. 

I have added a few comments, regarding the const keyword as it is used somewhere, but not everywhere.

include/ur_client_library/comm/tcp_socket.h
fmauch: That sounds reasonable.

fmauch: I've added it to the `setup` method and added a deprecation notice. Due to being implemented as a default parameter value I couldn't properly overload in order to keep the old and current API stable. Hence, I added a manual conflict resolution to choose the value set through the deprecated API with an additional deprecation notice.

fmauch:
@urmahp If you could give it one more review, that would be great. I didn't clean up the commits since this seems to confuse coverage.io quite a lot. I'll do that right before merging.

tests/test_producer.cpp
```diff
@@ -29,7 +29,9 @@
 // -- END LICENSE BLOCK ------------------------------------------------
 
 #include <gtest/gtest.h>
+#include <chrono>
 #include <condition_variable>
+#include <ratio>
```
urmahp, line 34: Is this used?

tests/test_tcp_socket.cpp
```diff
@@ -29,8 +29,14 @@
 // -- END LICENSE BLOCK ------------------------------------------------
 
 #include <gtest/gtest.h>
+#include <chrono>
 #include <condition_variable>
+#include <cstddef>
+#include <ratio>
```
urmahp, line 35: Is this used?

urmahp (approved):
Overall it looks very good. I only have a small comment regarding include of ratio in the test, which I believe is not used. 

Other than that I feel fine with merging this.

tests/test_producer.cpp
fmauch: thanks for spotting this, I removed it.

PR #175, opened 2023-09-04
Use pre-commit for clang-format

fmauch:
It has bothered be for quite a while that we have so many "code formatting" commit inside our PRs. Using pre-commit enables us to check formatting **before** we actually push something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,6 +381,19 @@ int main(int argc, char* argv[])
 }
 ```
 
+## Contributor Guidelines
+* This repo supports [pre-commit](https://pre-commit.com/) e.g. for automatic code formatting. TLDR:
+  This will prevent you from committing falsely formatted code:
+  ```
+  pipx install pre-commit
+  pre-commit install
+  ```
+* Succeeding pipelines are a must on Pull Requests (unless there is a reason, e.g. when there have
+been upstream changes).
+* We try to increase and keep our code coverage high, so PRs with new
+features should also have tests covering them.
+* Parameters of public methods must all be documented.
+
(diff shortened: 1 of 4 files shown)


urmahp (commented):
LGTM, but before merging I think we should add some documentation on how to use the pre-commit when you are developing, similar to what we have in the [ROS2 Driver](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver#contributor-guidelines)

fmauch:
@urmahp I've updated things  a bit together with a contributor guideline.

README.md
```diff
@@ -381,6 +381,19 @@ int main(int argc, char* argv[])
 }
 ```
 
+## Contributor Guidelines
+* This repo supports [pre-commit](https://pre-commit.com/) e.g. for automatic code formatting. TLDR:
+  This will prevent you from committing falsely formatted code:
+  ```
+  pipx install pre-commit
```
urmahp, line 388: I think we should tell them which clang format to install as well. 
```
```suggestion
  pipx install pre-commit 
  sudo apt install clang-format-14
```

urmahp (approved):
LGTM. Just added small suggestion regarding installing clang-format-14

README.md
fmauch: The clang-format version is specified in the pre-commit job. That's the main reason I switched to pre-commit: clang-format is installed in a python virtualenv with the respective version, so developers (and CI maintainers) don't have to manually install a specific version.

urmahp: I wasn't aware of that, nice solution :)

PR #176, opened 2023-09-04
Make tcp_server retry binding the socket

fmauch:
It sometimes happens that binding the socket fails since it is already in use (e.g. when restarting the driver too fast). This commit should retry binding instead of simply throwing an exception

### DIFF
--- a/include/ur_client_library/comm/tcp_server.h
+++ b/include/ur_client_library/comm/tcp_server.h
@@ -35,6 +35,7 @@
 #include <unistd.h>
 
 #include <atomic>
+#include <chrono>
 #include <functional>
 #include <thread>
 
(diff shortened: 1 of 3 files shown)


src/comm/tcp_server.cpp
```diff
@@ -31,6 +31,7 @@
 
 #include <iostream>
 
+#include <ratio>
```
urmahp, line 34: Is this used?

include/ur_client_library/comm/tcp_server.h
```diff
@@ -60,7 +61,8 @@ class TCPServer
 {
 public:
   TCPServer() = delete;
-  TCPServer(const int port);
+  TCPServer(const int port, const size_t max_num_tries = 0,
```
urmahp, line 64: I think we should add a some documentation for the parameters.

urmahp (approved):
I think it is a good improvement and overall it looks good to me. I just have two small comments.

src/comm/tcp_server.cpp
fmauch: thanks for catching this.

fmauch:
Holding for after #177 is merged.

PR #178, opened 2023-09-11
Added watchdog configuration for the reverse socket

urmahp:
This enables the possibility to configure the read timeout for the reverse socket running in the external control script

Separated control modes into realtime and non realtime to have different possible configurations for the watchdog

Added test to verify the changes

Updated test and examples to coincide with the new changes


src/ur/ur_driver.cpp
```diff
@@ -583,6 +589,61 @@ std::vector<std::string> UrDriver::getRTDEOutputRecipe()
...
+      return step_time;
+    }
+    else
+    {
+      return watchdog.timeout_;
+    }
+  }
+  else if (comm::ControlModeTypes::is_control_mode_realtime(control_mode))
+  {
+    float max_realtime_timeout = 1.0;
```
fmauch, line 619: While I agree with a mechanism to prevent users from having large timeouts on rt-commanding setting this here to 1 second seems a bit arbitrary. However, that's just a minor comment.

src/ur/ur_driver.cpp
```diff
@@ -259,9 +259,11 @@ std::unique_ptr<rtde_interface::DataPackage> urcl::UrDriver::getDataPackage()
   return rtde_client_->getDataPackage(timeout);
 }
 
-bool UrDriver::writeJointCommand(const vector6d_t& values, const comm::ControlMode control_mode)
+bool UrDriver::writeJointCommand(const vector6d_t& values, const comm::ControlMode control_mode,
+                                 const Watchdog& watchdog)
 {
-  return reverse_interface_->write(&values, control_mode);
+  const float read_timeout = verifyWatchdogTimeout(watchdog, control_mode, step_time_);
```
fmauch, line 265: Instead of doing this everywhere we could do this inside the reverse_interface inside the `write()` function, right?

urmahp: Yes we could do that, but there are still 3 write functions within the reverse interface, so it wouldn't make a big difference whether we do it in the UrDriver or within the reverse interface.

src/ur/ur_driver.cpp
urmahp: Yes it is an arbitrary timeout and I don't have a good argument for the 1 second, other than I don't think we should make it to large. So I am open for suggestions.

examples/spline_example.cpp
```diff
@@ -212,7 +210,7 @@ int main(int argc, char* argv[])
 
   bool ret = false;
   URCL_LOG_INFO("Switch to Forward mode");
-  ret = g_my_driver->writeJointCommand(vector6d_t(), comm::ControlMode::MODE_FORWARD);
+  ret = g_my_driver->writeTrajectoryControlMessage(control::TrajectoryControlMessage::TRAJECTORY_NOOP);
```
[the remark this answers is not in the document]
urmahp: I changed it to have the default timeout from the trajectory forward instead of setting the timeout manually.

include/ur_client_library/ur/watchdog.h
```diff
@@ -0,0 +1,69 @@
...
+
+#pragma once
+
+namespace urcl
+{
+/*!
+ * \brief Watchdog class containing a timeout configuration
+ *
+ * This watchdog is used to configure the read timeout for the reverse socket running in the external control script.
+ * The read timeout is the number of seconds until the read action times out (float). A timeout of 0 or negative number
```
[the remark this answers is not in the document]
urrsk: The reason why I have chosen to avoid a timer object is to try to make the calls as stateless as possible.
We chose this solution of having a argument to the write function of this type, instead of having a separate set function to a timer object, as we still need to handle edge cases and have some limits on the range.
I think this is specially doable if you mix the the different modes on the fly

examples/spline_example.cpp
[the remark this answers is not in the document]
fmauch: Yes, what I meant is: Does `URCL_LOG_INFO("Switch to Forward mode");` really make sense as an output?

include/ur_client_library/ur/watchdog.h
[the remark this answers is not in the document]
fmauch: Hm, I still cannot really follow. Why is a `Watchdog` object more stateless than an `std::chrono::milliseconds` object?

urmahp:
I have updated the PR based on your comments.

examples/spline_example.cpp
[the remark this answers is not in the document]
urmahp: Ahh I missed that, no that doesn't really make sense

include/ur_client_library/ur/watchdog.h
```diff
@@ -0,0 +1,92 @@
...
+/*!
+ * \brief Watchdog class containing a timeout configuration
+ *
+ * This watchdog is used to configure the read timeout for the reverse socket running in the external control script.
+ * The read timeout is the number of milliseconds until the read action times out. A timeout of 0 or negative number
+ * indicates that the function should not return until a read is completed, this will make the read function blocking on
+ * the robot. This can be set using the function off().
+ *
+ */
+class Watchdog
```
fmauch, line 47: While writing my previous comments about `std::chrono` I think I didn't realize what really bugged be: The `Watchdog` class isn't really a watchdog. When I read "watchdog" I expect an object that has a timeout, actually "watches" some process and raises something if that timeout is exceeded.

However, this class is merely a timeout itself with a couple of fancy construction functions. We could have the timeout as a single `std::chrono` object and pass that down to the write functions. The constructor methods could be simple macros e.g. `#define READ_TIMEOUT_OFF std::chrono::milliseconds(0)`. I'm not saying, we should, but we could without changing much.

The `verifyWatchdogTimeout()` method wouldn't loose much from its benefit if it was just a standalone helper function that takes the control_mode, step_time and timeout as arguments. Since we are calling this in each write method of the reverse interface manually, this wouldn't actually be much worse than the current approach.

I'm not sure whether I would vote for removing the class, but if we keep it I would like to propose a more meaningful name for the class. For example:
- `ControlTimeout`
- `ControlReadTimeout`
- `ControlTimeoutValidator`
- `ControlReadTimeoutValidator`
- `TimeoutValidator`

I think my favorite would be `ControlReadTimeout` since that best describes the use case.

urrsk: The reason why I have suggested this class/type is to avoid people use zero as an argument for the time and expected that this is the smallest timeout value where it in reality completely turn off and make the socket command a blocking call.
While still support that it could be a blocking call waiting for a response when the robot e.g. are standing still.

I agree regarding the naming, maybe just call it Timeout which is what it is.

fmauch: Agreed on `RobotReceiveTimeout`

urrsk:
We should also add a test, that verifies that the reverse interface timeout as well with in the expected time.

urmahp:
> We should also add a test, that verifies that the reverse interface timeout as well with in the expected time.

As the timeout is configured in the script the reverse interface will not timeout, the script will timeout. We could add a test that verifies that the script times out, but that would be a test under the UrDriver class, as it would require robot communication.

urrsk:
> > We should also add a test, that verifies that the reverse interface timeout as well with in the expected time.
> 
> As the timeout is configured in the script the reverse interface will not timeout, the script will timeout. We could add a test that verifies that the script times out, but that would be a test under the UrDriver class, as it would require robot communication.

True. And yes that is a good idea

src/control/reverse_interface.cpp
```diff
@@ -55,8 +61,19 @@ bool ReverseInterface::write(const vector6d_t* positions, const comm::ControlMod
   uint8_t buffer[sizeof(int32_t) * MAX_MESSAGE_LENGTH];
   uint8_t* b_pos = buffer;
 
-  // The first element is always the keepalive signal.
-  int32_t val = htobe32(keepalive_count_);
+  int read_timeout = robot_receive_timeout.verifyRobotReceiveTimeout(comm::ControlMode::MODE_FREEDRIVE, step_time_);
```
fmauch, line 64: Is there a reason we're not using the actual `control_mode` here?

src/ur/robot_receive_timeout.cpp
```diff
@@ -0,0 +1,116 @@
...
+RobotReceiveTimeout RobotReceiveTimeout::off()
+{
+  RobotReceiveTimeout robot_receive_timeout(std::chrono::milliseconds(0));
+  return robot_receive_timeout;
+}
+
+int RobotReceiveTimeout::verifyRobotReceiveTimeout(const comm::ControlMode control_mode,
+                                                   const std::chrono::milliseconds step_time) const
+{
+  // Convert timeout to float
```
fmauch, line 67: I guess that comment can be removed.
```suggestion
```

src/ur/robot_receive_timeout.cpp
```diff
@@ -0,0 +1,116 @@
...
+  if (comm::ControlModeTypes::is_control_mode_non_realtime(control_mode))
+  {
+    if (timeout_ < step_time && timeout_ > std::chrono::milliseconds(0))
+    {
+      std::stringstream ss;
+      ss << "Robot receive timeout " << timeout_.count() << " is below the step time " << step_time.count()
+         << ". It will be reset to the step time.";
```
fmauch, line 74:
```suggestion
      ss << "Robot receive timeout " << timeout_.count() << "ms is below the step time " << step_time.count()
         << "ms. It will be reset to the step time.";
```

src/ur/robot_receive_timeout.cpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+  else if (comm::ControlModeTypes::is_control_mode_realtime(control_mode))
+  {
+    const std::chrono::milliseconds max_realtime_timeout = std::chrono::milliseconds(1000);
+    if (timeout_ < step_time)
+    {
+      std::stringstream ss;
+      ss << "Realtime read timeout " << timeout_.count() << " is below the step time " << step_time.count()
+         << ". It will be reset to the step time.";
```
fmauch, line 90:
```suggestion
      ss << "Realtime read timeout " << timeout_.count() << "ms is below the step time " << step_time.count()
         << "ms. It will be reset to the step time.";
```

src/ur/robot_receive_timeout.cpp
```diff
@@ -0,0 +1,116 @@
...
+      URCL_LOG_ERROR(ss.str().c_str());
+      return step_time.count();
+    }
+    else if (timeout_ > max_realtime_timeout)
+    {
+      std::stringstream ss;
+      ss << "Robot receive timeout " << timeout_.count()
+         << " is above the maximum allowed timeout for realtime commands " << max_realtime_timeout.count()
+         << ". It will be reset to the maximum allowed timeout.";
```
fmauch, line 99:
```suggestion
         << "ms is above the maximum allowed timeout for realtime commands " << max_realtime_timeout.count()
         << "ms. It will be reset to the maximum allowed timeout.";
```

src/ur/robot_receive_timeout.cpp
```diff
@@ -0,0 +1,116 @@
...
+      return step_time.count();
+    }
+    else
+    {
+      return timeout_.count();
+    }
+  }
+  else if (comm::ControlModeTypes::is_control_mode_realtime(control_mode))
+  {
+    const std::chrono::milliseconds max_realtime_timeout = std::chrono::milliseconds(1000);
```
fmauch, line 85: Should we wrap this into an explicit constant `MAX_RT_RECEIVE_TIMEOUT_MS`? Doing it this way it is kind of a hidden magic number. Also: should that maybe be a bit lower? I would have propoped something in the range of 100ms, maybe 200ms.

tests/test_control_mode.cpp
```diff
@@ -0,0 +1,87 @@
...
+TEST(control_mode, is_control_mode_realtime_with_non_realtime_control_modes)
+{
+  for (unsigned int i = 0; i < comm::ControlModeTypes::NON_REALTIME_CONTROL_MODES.size(); ++i)
+  {
+    EXPECT_FALSE(
+        comm::ControlModeTypes::is_control_mode_realtime(comm::ControlModeTypes::NON_REALTIME_CONTROL_MODES[i]));
+  }
+
+  // control modes that are neither realtime or not realtime, but they should also return false
+  EXPECT_FALSE(comm::ControlModeTypes::is_control_mode_realtime(comm::ControlMode::MODE_STOPPED));
```
fmauch, line 54: For testing purposes it might make sense to add the third group, so we can iterate over them, as well. Then, we can additionally check that the size of all three groups summed matches the total number of control modes. This way we ensure that adding a control mode later requires grouping that or otherwise the test fails.

fmauch (commented):
This is looking really good and it reads much better with the RobotReceiveTimeout :-)

@urmahp a quick response to my comments would be nice, then I think we are good to merge this, as most of them is mainly aesthetics.

src/control/reverse_interface.cpp
urmahp: No, that was a copy paste mistake by me...

src/ur/robot_receive_timeout.cpp
urmahp: That's a good idea, I will vote for 200ms.

tests/test_control_mode.cpp
urmahp: That's a good idea, I will add a group and a test.

urmahp:
I have updated the PR in relation to the comments and added a better test coverage of the changes.

fmauch (approved):
This looks good to me! I'll deploy it on my test system and merge if everything goes fine on the real robot.

PR #180, opened 2023-09-18
Release 1.3.4

fmauch:
Release PR for next release.


fmauch:
Before actually bloom-releasing this we should update the drivers to use the timeout instead of the keepalive count as this raises an error in startup.

Thinking about this: Shouldn't we make the deprecation notice a warning?

urmahp (approved):
LGTM and I also think we should make the deprecation notice a warning as mentioned in the other PR.

fmauch:
Waiting for #183. Sorry I didn't realize the warning was in two places...

PR #182, opened 2023-09-22
Make depreaction warning for keepalive_counter a warning instead of error

fmauch:
As suggested in #180 

I overlooked that when reviewing #178, but noticed that when starting a driver built against the newest version.


urmahp (approved):
I agree with making it a warning.

PR #184, opened 2023-09-26
Added check to ensure that the targets are reachable within the robot…

urmahp:
…s limits

This will break the test_joint_trajectory_position_interface test in the ROS Driver as part of the test is designed to activate the speed scaling by sending unreachable targets. These changes therefore require that we update the test in the ROS Driver.


resources/external_control.urscript
```diff
@@ -113,11 +146,21 @@ thread servoThread():
       end
 
       q = extrapolate()
-      servoj(q, t=steptime, {{SERVO_J_REPLACE}})
+      if targetWithinLimits(q, steptime):
+        servoj(q, t=steptime, {{SERVO_J_REPLACE}})
+        q_act = get_actual_joint_positions()
```
fmauch, line 151:
```suggestion
```

resources/external_control.urscript
```diff
@@ -113,11 +146,21 @@ thread servoThread():
...
+      else:
+        terminateProgram()
+      end
 
     elif state == SERVO_RUNNING:
       extrapolate_count = 0
-      servoj(q, t=steptime, {{SERVO_J_REPLACE}})
+      if targetWithinLimits(q, steptime):
+        servoj(q, t=steptime, {{SERVO_J_REPLACE}})
+        q_act = get_actual_joint_positions()
```
fmauch, line 160:
```suggestion
```

resources/external_control.urscript
```diff
@@ -113,11 +146,21 @@ thread servoThread():
...
+      else:
+        terminateProgram()
```
fmauch, line 162: I wonder what would be the benefit of actually quitting here? We could instead just ignore the command.

fmauch:
While this is in a state where everything seems to do its job as intended, it's actually not as simple as that. Since we use `servoj` for executing trajectories in the position-based case, there is some delay between the moment where we call servoj and servoj actually starting the motion. Thus, simply looking at the current positions and the ones coming in as new targets will not necessarily reflect the actual velocities. I'll try to add a test showing this.

fmauch:
I've updated this so that it doesn't terminate the program but instead ignores the commands as long as they are infeasible. With this, when I start this with a driver that always commands [0,0,0,0,0,0], the robot doesn't move until we send a target that is close to its current position.

I used a non-blocking popup as we discussed, but thinking about this, a blocking popup might actually be more feasible, don't you agree @urrsk ?

resources/external_control.urscript
```diff
@@ -77,6 +84,48 @@ thread_move = 0
...
+      if last_violation_popup_counter == 0:
+        textmsg(str)
+        popup(str, blocking=False)
+      elif last_violation_popup_counter * get_steptime() > 5.0:
+        # We want to have a log printed regularly. We are receiving motion commands that are not
+        # feasible. The user should have a chance to know about this.
+        textmsg(str)
+        last_violation_popup_counter = 0
+      end
+      last_violation_popup_counter = last_violation_popup_counter + 1
```
urrsk, line 111:
```suggestion
      if last_violation_popup_counter == 0:
        # Keep making a popup as long as the targets are too much off
        local str = str_cat(str_cat("Velocity ", velocity), str_cat(str_cat(" required to reach the received target ", step_end), str_cat(str_cat(" within ", time), " seconds is exceeding the joint velocity limits. Ignoring commands until a valid command is received.")))
        textmsg(str)
        popup(str, error=True, blocking=False)
      end
      # Report the error every 5 sec
      last_violation_popup_counter = last_violation_popup_counter + 1
      if last_violation_popup_counter * get_steptime() >= 5.0:
        last_violation_popup_counter = 0
      end
```
I think it is fine it is not blocking, as the alterative is more to terminating the program instead. Though let's ensure that the popup keeps coming if a continues stream of invalid targets keeps coming

urrsk (changes requested):

fmauch:
@urrsk Now everything is where I think I would like it to be and the tests are in line, as well. If you could take another look, that would be awesome. Or @VinDp.

resources/external_control.urscript
```diff
@@ -160,16 +223,23 @@ def cubicSplineRun(end_q, end_qd, time, is_last_point=False):
...
 def quinticSplineRun(end_q, end_qd, end_qdd, time, is_last_point=False):
   local str = str_cat(end_q, str_cat(end_qd, str_cat(end_qdd, time)))
   textmsg("Quintic spline arg: ", str)
 
-  # Zero time means zero length and therefore no motion to execute
-  if time > 0.0:
-    local start_q = get_target_joint_positions()
+  local start_q = get_target_joint_positions()
+  # Zero time means infinite velocity to reach the target and is therefore impossible
+  if time <= 0.0:
```
urrsk, line 237: I am afraid that some people are putting the robot's current position as the first node in the trajectory.
Maybe we could ignore this not with the time of zero if:
1. it is the first node in the path
1. it is very close to the robot's current position

resources/external_control.urscript
```diff
@@ -387,15 +462,15 @@ thread trajectoryThread():
...
         end
       end
     end
   end
   exit_critical
   stopj(STOPJ_ACCELERATION)
-  socket_send_int(TRAJECTORY_RESULT_SUCCESS, "trajectory_socket")
-  textmsg("Trajectory finished")
+  socket_send_int(trajectory_result, "trajectory_socket")
+  textmsg("Trajectory finished with result ", trajectory_result)
```
urrsk, line 473: Maybe convert the trajectory_result to a readable string before sending it to the log

urrsk (changes requested):

fmauch:
@urrsk I implemented the things discussed. I was a bit puzzled as I found out that when ignoring the first trajectory point, apparently the read-from-socket call times out. That's why I increased the timeout in this PR. However, that doesn't seem right to me. I checked that sending out the trajectory points only took a few microseconds in the test program, so the timeout could either result from

* The network delaying the package
* Delay between socket send command and package actually entering the network layer
* Socket read in the thread being at lower priority?

I quickly modified the `trajectoryThread()` in order to just print the points to a textmsg instead of executing them and that showed the same behavior. The first trajectory point was printed fine, the others timed out.

urrsk:
> @urrsk I implemented the things discussed. I was a bit puzzled as I found out that when ignoring the first trajectory point, apparently the read-from-socket call times out. That's why I increased the timeout in this PR. However, that doesn't seem right to me. I checked that sending out the trajectory points only took a few microseconds in the test program, so the timeout could either result from
> 
> * The network delaying the package
> * Delay between socket send command and package actually entering the network layer
> * Socket read in the thread being at lower priority?
> 
> I quickly modified the `trajectoryThread()` in order to just print the points to a textmsg instead of executing them and that showed the same behavior. The first trajectory point was printed fine, the others timed out.

Interesting, it seems we need to do a bit more investigation on this

fmauch:
@urrsk as discussed I use the higher timeout when the robot isn't moving, yet only. Could you please have another look?

fmauch:
I'll investigate why the noetic tests fail. A quick look didn't make it clear to me.

.github/workflows/industrial-ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         ROS_DISTRO:
-          - {NAME: noetic, DOWNSTREAM_WORKSPACE: "github:UniversalRobots/Universal_Robots_ROS_Driver#master https://raw.githubusercontent.com/UniversalRobots/Universal_Robots_ROS_Driver/master/.noetic.rosinstall"}
+          - {NAME: noetic, DOWNSTREAM_WORKSPACE: "github:fmauch/Universal_Robots_ROS_Driver#fix_trajectory_forward_test https://raw.githubusercontent.com/UniversalRobots/Universal_Robots_ROS_Driver/master/.noetic.rosinstall"}
```
fmauch, line 12:
```suggestion
          - {NAME: noetic, DOWNSTREAM_WORKSPACE: "github:UniversalRobots/Universal_Robots_ROS_Driver#master https://raw.githubusercontent.com/UniversalRobots/Universal_Robots_ROS_Driver/master/.noetic.rosinstall"}
```

fmauch (changes requested):

fmauch:
> I'll investigate why the noetic tests fail. A quick look didn't make it clear to me.

Found it. The actual test wasn't working properly and this got uncovered by this PR. Necessary changes are being implemented in https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/721

fmauch (approved):
Approving so my change-request block goes away.

urrsk (approved):
I have mainly focused on the script code in my review

PR #186, opened 2023-09-29
This PR improves the primary interface support in the client library

urmahp:
A big acknowledgement should be given to @fmauch and @urmarp as they have build the foundation for this PR in #26 and #137.

I have tried to address the most important issues with the primary interface as it is today based on the work from #26 and #137, but without including all the changes from the PRs. This means that I have added more feedback when you send script commands to the robot and we also report if there is an error in the script send to the robot. Furthermore information received over the primary interface is logged in the shell. 

Unfortunately I haven't had the time to add tests for the changes and this should definitely be done before merging.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,10 +25,16 @@ add_library(urcl SHARED
     src/control/trajectory_point_interface.cpp
     src/control/script_command_interface.cpp
     src/primary/primary_package.cpp
+    src/primary/program_state_message.cpp
     src/primary/robot_message.cpp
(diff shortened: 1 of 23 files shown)


michaelaeriksen:
Any update on this?

michaelaeriksen:
Any plans to merge this into main?

urrsk:
Thanks for asking. Yes, but as you see we still missing some test to cover this work before it is ready to be merge.
@michaelaeriksen you are also welcome help making the unit tests.

PR #188, opened 2023-10-18
Remove duplicated entry in clang-format file

fmauch:
This has been hanging around for a while, but some tools seem to be complaining about this.


RobertWilbrandt (approved):
Looks good

PR #189, opened 2023-11-16
Add header guard to datatypes.h

fmauch:


RobertWilbrandt (approved):
Good change that prevents simple user errors. I just realized that we are not 100% consistent with pragma once vs. header guard in our header files, maybe something to keep in mind in the future.

Issue #194, opened 2024-02-21
Interpolation on robot gives discontinuous motion when speed scaling is active


urmahp:
When using the spline interpolation on the robot with speed scaling active, there is a discontinuous motion. The reason for this is that the target joint velocity is used when calculating the next spline in the trajectory see [here](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/resources/external_control.urscript#L153) and [here](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/resources/external_control.urscript#L172), but when speed scaling is active the target velocity is no longer the same velocity as the velocity in the trajectory. This means that the spline will be wrongly calculated and this results in a discontinuous motion, because the robot slows down before speeding up in the beginning of each spline within the trajectory.

There is also another issue with the spline interpolation on the robot. Within the interpolation speedj command is called with the target joint velocities calculated from the spline interpolation and an acceleration of 1000. This large acceleration can make the joints noisy, due to the acceleration spikes every time speedj is called.

PR #195, opened 2024-02-21
Changed spline interpolation to use the last commanded joint velocity…

urmahp:
… when calculating the next spline in the trajectory

Also limited the acceleration to the maximum joint acceleration of the target accelerations calculated from the spline. But still added a constant acceleration when slowing down is active, as the acceleration in the trajectory no longer can be used.

This will fix the issue with discontinuous motion when speed scaling is active

This will fix issue #194

### DIFF
--- a/resources/external_control.urscript
+++ b/resources/external_control.urscript
@@ -68,8 +68,8 @@ global extrapolate_count = 0
 global extrapolate_max_count = 0
 global control_mode = MODE_UNINITIALIZED
 global trajectory_points_left = 0
-global last_spline_qdd = [0, 0, 0, 0, 0, 0]
-global last_spline_qd = [0, 0, 0, 0, 0, 0]
+global spline_qdd = [0, 0, 0, 0, 0, 0]
+global spline_qd = [0, 0, 0, 0, 0, 0]
 global tool_contact_running = False
 
 # Global thread variables
@@ -151,7 +151,7 @@ def cubicSplineRun(end_q, end_qd, time, is_last_point=False):
   # Zero time means zero length and therefore no motion to execute
   if time > 0.0:
     local start_q = get_target_joint_positions()
-    local start_qd = last_spline_qd
+    local start_qd = spline_qd
 
     # Coefficients0 is not included, since we do not need to calculate the position
     local coefficients1 = start_qd
@@ -170,8 +170,8 @@ def quinticSplineRun(end_q, end_qd, end_qdd, time, is_last_point=False):
   # Zero time means zero length and therefore no motion to execute
   if time > 0.0:
     local start_q = get_target_joint_positions()
-    local start_qd = last_spline_qd
-    local start_qdd = last_spline_qdd
+    local start_qd = spline_qd
+    local start_qdd = spline_qdd
 
     # Pre-calculate coefficients
     local TIME2 = pow(time, 2)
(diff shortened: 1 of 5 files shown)


resources/external_control.urscript
```diff
@@ -298,6 +305,31 @@ def checkSlowDownRequired(x, qd, max_speed, max_deceleration):
...
+def list_max_norm(list):
+  # ensure we have something to iterate over
+  local length = get_list_length(list)
+  if  length == 0:
+      popup("Getting the maximum of an empty list is impossible in list_max().", error = True, blocking = True)
+      halt
+  end
+
+  # search for maximum
+  local i = 0
```
urrsk, line 321: To improve efficiency
```suggestion
  local i = 1
```

resources/external_control.urscript
```diff
@@ -298,6 +305,31 @@ def checkSlowDownRequired(x, qd, max_speed, max_deceleration):
...
 
+###
+# @brief Find the maximum value in a list the list must be of non-zero length and contain numbers
+# @param list array list
+###
+def list_max_norm(list):
+  # ensure we have something to iterate over
+  local length = get_list_length(list)
+  if  length == 0:
+      popup("Getting the maximum of an empty list is impossible in list_max().", error = True, blocking = True)
```
urrsk, line 316: Not sure that this is the best way to handle the case of receiving an empty list.
Maybe ignore the trajectory instead (with an error message) and let the driver continue to be ready to receive the next command.

resources/external_control.urscript
```diff
@@ -260,14 +261,20 @@ def jointSplineRun(coefficients1, coefficients2, coefficients3, coefficients4, c
...
   end
 
-  jointSplineStep(coefficients1, coefficients2, coefficients3, coefficients4, coefficients5, splineTotalTravelTime, timeLeftToTravel, scaling_factor)
+  jointSplineStep(coefficients1, coefficients2, coefficients3, coefficients4, coefficients5, splineTotalTravelTime, timeLeftToTravel, scaling_factor, is_slowing_down)
 end
 
-def jointSplineStep(coefficients1, coefficients2, coefficients3, coefficients4, coefficients5, splineTimerTraveled, timestep, scaling_factor):
-  local qd = coefficients1 + 2.0 * splineTimerTraveled * coefficients2 + 3.0 * pow(splineTimerTraveled, 2) * coefficients3 + 4.0 * pow(splineTimerTraveled, 3) * coefficients4 + 5.0 * pow(splineTimerTraveled, 4) * coefficients5
+def jointSplineStep(coefficients1, coefficients2, coefficients3, coefficients4, coefficients5, splineTimerTraveled, timestep, scaling_factor, is_slowing_down=False):
+  last_spline_qd = coefficients1 + 2.0 * splineTimerTraveled * coefficients2 + 3.0 * pow(splineTimerTraveled, 2) * coefficients3 + 4.0 * pow(splineTimerTraveled, 3) * coefficients4 + 5.0 * pow(splineTimerTraveled, 4) * coefficients5
```
urrsk, line 268: Try to find a better name, as it only is the last_spline_qd next time!

resources/external_control.urscript
```diff
@@ -260,14 +261,20 @@ def jointSplineRun(coefficients1, coefficients2, coefficients3, coefficients4, c
...
   last_spline_qdd = 2.0 * coefficients2 + 6.0 * splineTimerTraveled * coefficients3 + 12.0 * pow(splineTimerTraveled, 2) * coefficients4 + 20.0 * pow(splineTimerTraveled, 3) * coefficients5
-  qd = qd * scaling_factor
-  speedj(qd, 1000, timestep)
+
+  if is_slowing_down:
+    last_spline_qd = last_spline_qd * scaling_factor
+    speedj(last_spline_qd, 15.0, timestep)
```
urrsk, line 273: Avoid hardcoded duplicates constants. Use a variable instead

resources/external_control.urscript
```diff
@@ -260,14 +261,20 @@ def jointSplineRun(coefficients1, coefficients2, coefficients3, coefficients4, c
...
+  else:
+    max_qdd = list_max_norm(last_spline_qdd)
```
urrsk, line 275: Suggest that we calculate the qdd argument for speedj more simple way; by substracting the last qd with the current calculated qd.
In practice it will not have any impact as we are doing it in very small intervals/steps

urrsk (changes requested):
Good suggestion. Lets take it through an extra iterations and fix the test as well

tests/test_spline_interpolation.cpp
```diff
@@ -388,27 +472,36 @@ TEST_F(SplineInterpolationTest, cubic_spline_with_end_point_velocity)
...
+  ASSERT_NEAR(spline_travel_time, s_time.back(), 1e-5);
+
+  writeTrajectoryToFile("../test_artifacts/cubic_spline_with_end_point_velocity.csv", time_vec, expected_positions,
+                        actual_positions, actual_velocities, actual_acc, speed_scaling_vec, spline_time);
 }
 
-TEST_F(SplineInterpolationTest, quintic_spline_with_end_point_velocity)
+TEST_F(SplineInterpolationTest, quintic_spline_with_end_point_velocity_with_speedscaling)
 {
+  // Set target speed scaling to 100% as one test change this value
```
urmahp, line 486: This is the desctructor comment, should probably be something like
```suggestion
  // Set speed scaling to 25% to test interpolation with speed scaling active
```

tests/test_spline_interpolation.cpp
```diff
@@ -459,8 +558,46 @@ TEST_F(SplineInterpolationTest, quintic_spline_with_end_point_velocity)
...
+      for (unsigned int i = 0; i < last_joint_acc.size(); ++i)
+      {
+        change_acc[i] = joint_acc[i] - last_joint_acc[i];
+
+        if (!nearly_equal(change_acc[i], 0.0, EPS_ACC_CHANGE) && !nearly_equal(last_change_acc[i], 0.0, EPS_ACC_CHANGE))
+        {
+          // Acceleration should only increase or be constant within one scaled timescale.
+          // It should not fluctuate to zero or overshoot
+          EXPECT_EQ(sign(last_change_acc[i]), sign(change_acc[i]))
+              << " acceleration change direction doing "
```
urmahp, line 592: I think a debug message wasn't fully deleted here.

urmahp (commented):
Overall it looks good and I am fine with merging this.

tests/test_spline_interpolation.cpp
urrsk: This is intended. The debug message is only printed out if the EXPECT_EQ fails.

urrsk (approved):

PR #201, opened 2024-05-23
URSim: Use a program folder per model

### LINKED ISSUE #190: [start_ursim] Sharing of urcap and program storage in series can lead to errors
`start_ursim.sh` defaults to using `${PERSISTENT_BASE}/${ROBOT_SERIES}` as base for urcap and program storage. This means that the same storage is shared between different robots of a series which can lead to errors:
1. `rm -rf ~/.ursim/e-series`
2. `ros2 run ur_client_library start_ursim.sh -m ur5e`
    This will create the relevant files and a safety configuration with 191 deg/s for each joint.
3. `ros2 run ur_client_library start_ursim.sh -m ur10e`
    This shares the safety configuration, leading to safety errors as soon as you go to the installation tab, as the joint limits for the first two joints are only allowed to go up to 131 deg/s for this arm.

This should be fixed to allow for easier switching between simulated robots.

fmauch:
Otherwise we switch installation between different models all the time.

This closes #190

### DIFF
--- a/scripts/start_ursim.sh
+++ b/scripts/start_ursim.sh
@@ -54,8 +54,6 @@ help()
 ROBOT_MODEL=UR5
 ROBOT_SERIES=e-series
 URSIM_VERSION=latest
-URCAP_STORAGE="${PERSISTENT_BASE}/${ROBOT_SERIES}/urcaps"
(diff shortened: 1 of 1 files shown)


VinDp (approved):
Tested it: works nicely, good fix.

PR #206, opened 2024-07-02
Moved reset of speed slider to correct teardown function, so that it …

URJala:
…resets between each test.


fmauch:
@URJala could you please rebase that and revert the re-formatting that is throwing the error? Apart from that this looks fine, thank you!

URJala:
Yes, sorry, it should be good now :)

fmauch (approved):

PR #218, opened 2024-10-17
Allow resetting the RTDE client

fmauch:
With some control modes such as the trajectory passthrough controller it is actually possible to run the communication at much lower frequencies.

However, so far we always initialize the `RTDEClient` with the maximum frequency that is possible on the respective model.

This PR adds a function to `UrDriver` to reset the `RTDEClient` with another communication frequency.

## ToDo

- [x] Implementation
- [x] Write tests
- [x] Write documentation

### DIFF
--- a/doc/architecture.rst
+++ b/doc/architecture.rst
@@ -63,6 +63,11 @@ sure to
 * run it from the package's main folder, as for simplicity reasons it doesn't use any sophisticated
   method to locate the required files.
 
+.. note::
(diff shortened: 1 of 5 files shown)


urfeex:
Note about CI: The ROS tests onto the main repo are currently expected to fail due to the ongoing changes of ros2_control. As long as the tests against testing are fine, we should be fine.

urrsk (approved):
Looks good

PR #223, opened 2024-11-13
CI: Fix flaky example runs

urfeex:
The example runs seem to fail quite often. This PR aims at fixing that.

### DIFF
--- a/examples/dashboard_example.cpp
+++ b/examples/dashboard_example.cpp
@@ -36,6 +36,7 @@
 #include <iostream>
 #include <memory>
 #include <thread>
+#include <unistd.h>
 
 using namespace urcl;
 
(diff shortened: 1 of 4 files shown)


examples/dashboard_example.cpp
```diff
@@ -96,6 +97,8 @@ int main(int argc, char* argv[])
     return 1;
   }
 
+  sleep(1);
```
urrsk, line 100: Do we need this sleep?

urfeex: That's effectively what makes this non-flaky. Without that, the play command seems to be failing quite regularly in ci.

SarveshMalladi (approved):

PR #230, opened 2024-12-10
Enable force mode compatibility with various move types

urrsk:
This should make force mode work with various of move types. 
With this commit, the joint positions used for checks and interpolations is now done without force mode modifies.

This commit relates to #138


urfeex (approved):
This does make sense and I manually checked that it works as expected.

PR #237, opened 2024-12-12
Remove the not regarding MIT license

urrsk:
We no longer have the docker of ursim inside the client library, which originally gave us this dependency to the MIT license


urfeex (approved):
Good catch!

PR #242, opened 2025-01-10
Add instruction executor for high-level robot control

urfeex:
With the trajectory point interface there is a method of executing motions with only the urcl present as shown in https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/examples/trajectory_point_interface.cpp

However, usage is a bit cumbersome, we might want to support a one-liner for executing ptp motions, instead.

This PR adds the `InstructionExecutor`, a high-level module that uses existing functionality to provide simple-to-use interfaces, currently to execute MoveJ, MoveL and sequences of the two. In future, we might extend that to other instructions such as other motion types, but possibly also other functionality, hence the more general name.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -7,9 +7,8 @@ project(ur_driver_examples)
 # # Check C++11 support / enable global pedantic and Wall
 # #
 include(DefineCXX17CompilerFlag)
-#DEFINE_CXX_17_COMPILER_FLAG(CXX17_FLAG)
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++20")
+DEFINE_CXX_17_COMPILER_FLAG(CXX17_FLAG)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic")
 
 add_executable(driver_example
   full_driver.cpp)
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -7,8 +7,9 @@ project(ur_driver_examples)
 # # Check C++11 support / enable global pedantic and Wall
 # #
 include(DefineCXX17CompilerFlag)
-DEFINE_CXX_17_COMPILER_FLAG(CXX17_FLAG)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic")
+#DEFINE_CXX_17_COMPILER_FLAG(CXX17_FLAG)
+#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++20")
 
 add_executable(driver_example
   full_driver.cpp)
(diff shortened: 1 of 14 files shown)


examples/instruction_executor.cpp
```diff
@@ -0,0 +1,135 @@
...
+  if (argc > 1)
+  {
+    robot_ip = std::string(argv[1]);
+  }
+
+  // --------------- INITIALIZATION BEGIN -------------------
+  // Making the robot ready for the program by:
+  // Connect the robot Dashboard
+  g_my_dashboard.reset(new urcl::DashboardClient(robot_ip));
+  //  if (!g_my_dashboard->connect())
```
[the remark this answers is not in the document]
urfeex: That will be done for all examples on a follow-up PR.

examples/CMakeLists.txt
```diff
@@ -7,8 +7,9 @@ project(ur_driver_examples)
 # # Check C++11 support / enable global pedantic and Wall
 # #
 include(DefineCXX17CompilerFlag)
-DEFINE_CXX_17_COMPILER_FLAG(CXX17_FLAG)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic")
+#DEFINE_CXX_17_COMPILER_FLAG(CXX17_FLAG)
```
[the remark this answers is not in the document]
urfeex: has been done on the master branch.

src/control/trajectory_point_interface.cpp
```diff
@@ -197,8 +197,12 @@ void TrajectoryPointInterface::connectionCallback(const int filedescriptor)
 
 void TrajectoryPointInterface::disconnectionCallback(const int filedescriptor)
 {
-  URCL_LOG_DEBUG("Connection to trajectory interface dropped.", filedescriptor);
+  URCL_LOG_INFO("Connection to trajectory interface dropped.");
```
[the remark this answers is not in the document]
urfeex: That should be reverted, thanks.

tests/test_instruction_executor.cpp
```diff
@@ -170,26 +170,32 @@ TEST_F(InstructionExecutorTest, execute_movej_success)
   ASSERT_TRUE(executor_->moveJ({ -1.57, -1.57, 0, 0, 0, 0 }, 3.4, 3.1, 0));
   auto end = std::chrono::steady_clock::now();
   auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
-  ASSERT_NEAR(duration.count(), 1550, 100);
+  // With this parametrization execution should take about 1.4 seconds plus some overhead. We test
+  // time parametrization below with a motion time of 2.5 seconds, so this is the upper bound to
+  // distinguish between the two. This large range is necessary, as the actual overhead is not
+  // known.
+  ASSERT_GT(duration.count(), 1400);
```
SarveshMalladi, line 177: Shall we add some failures to test if the error codes match? 
EXCEPTIONS maybe?

examples/instruction_executor.cpp
```diff
@@ -0,0 +1,135 @@
...
+  //  {
+  //    URCL_LOG_ERROR("Could not connect to dashboard");
+  //    return 1;
+  //  }
+  //
```
SarveshMalladi, line 77: What about this?

urfeex: Thanks for the reminder. I put it back in.

tests/test_instruction_executor.cpp
urfeex: good point, thank you!

SarveshMalladi (approved):

PR #244, opened 2025-01-10
Modernize cmake

urfeex:
Modernize CMake of the client library with regards to

- Properly set C++17 flag using target_compile_features
- Properly include googletest e.g. using [CMakeFetchContent](https://google.github.io/googletest/quickstart-cmake.html#set-up-a-project)

These proposed changes should make this project easier to maintain and build in different environments.


urrsk (approved):
Nice improvement.
We could also consider if we could glob the test and examples to minimize the risk that we forget to add a test or an example to the CMakeLists.
Though to do it, we need to find a way to handle the "integration" tests. Maybe by a separate folder for them

urfeex:
> Nice improvement. We could also consider if we could glob the test and examples to minimize the risk that we forget to add a test or an example to the CMakeLists. Though to do it, we need to find a way to handle the "integration" tests. Maybe by a separate folder for them

Would you be fine with creating a separate issue for that? I think that's a valuable suggestion, but independent of this PR.

urrsk:
> > We could also consider if we could glob the test and examples to minimize the risk that we forget to add a test or an example to the CMakeLists.
> 
> `glob`-ing in `CMakeLists.txt` is a bit of a controversial topic. See [Is it better to specify source files with GLOB or each file individually in CMake?](https://stackoverflow.com/questions/1027247) fi. [This answer](https://stackoverflow.com/a/55290546) suggests a possible compromise, so you could still do it, but there's a bit more to it than just `glob`-ing everything.

Good point. Doing development, you need to know that you have to reload cmake to include new files as the CMakeLists.txt is not updated and therefore does not trigger it to update.

My main motivation is to ensure that our CI always will evaluate the test that exist in the test folders. It is so easy to forget adding to the cmake files, especially for externals that are contributing with PRs.

Therefore, I think it is good for test code. Though, I agree it is not a good approach for the application code

urfeex:
I'll go ahead and merge this. Let's move that discussion to a new issue.

PR #249, opened 2025-01-21
Do not throw exception in DashboardClient::sendRequest

urfeex:
I don't know why we changed it to throwing an exception. It doesn't make sense for this function.

### DIFF
--- a/tests/test_dashboard_client.cpp
+++ b/tests/test_dashboard_client.cpp
@@ -221,29 +221,25 @@ TEST_F(DashboardClientTest, connect_non_running_robot)
 
 TEST_F(DashboardClientTest, non_expected_result_returns_correctly)
 {
-  std::unique_ptr<DashboardClient> dashboard_client;
-  dashboard_client.reset(new DashboardClient("192.168.56.101"));
-  ASSERT_TRUE(dashboard_client->connect());
-  EXPECT_TRUE(dashboard_client->commandPowerOff());
+  ASSERT_TRUE(dashboard_client_->connect());
+  EXPECT_TRUE(dashboard_client_->commandPowerOff());
 
   // We will not get this answer
-  EXPECT_FALSE(dashboard_client->sendRequest("brake release", "non-existing-response"));
+  EXPECT_FALSE(dashboard_client_->sendRequest("brake release", "non-existing-response"));
 
   // A non-matching answer should throw an exception for this call
-  EXPECT_THROW(dashboard_client->sendRequestString("brake release", "non-existing-response"), UrException);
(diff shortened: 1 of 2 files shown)


tests/test_dashboard_client.cpp
```diff
@@ -219,6 +219,33 @@ TEST_F(DashboardClientTest, connect_non_running_robot)
   EXPECT_LT(elapsed, 2 * comm::TCPSocket::DEFAULT_RECONNECTION_TIME);
 }
 
+TEST_F(DashboardClientTest, non_expected_result_returns_correctly)
+{
+  std::unique_ptr<DashboardClient> dashboard_client;
+  dashboard_client.reset(new DashboardClient("192.168.56.101"));
```
urrsk, line 225:
```suggestion
  dashboard_client.reset(new DashboardClient(g_ROBOT_IP));
```

tests/test_dashboard_client.cpp
```diff
@@ -219,6 +219,33 @@ TEST_F(DashboardClientTest, connect_non_running_robot)
...
+
+  // Waiting for a non-matching answer should return false
+  // Internally we wait 100ms between each attempt, hence the 300ms wait time
+  EXPECT_FALSE(
+      dashboard_client->waitForReply("brake_release", "non-existing-response", std::chrono::milliseconds(300)));
+}
+
+TEST_F(DashboardClientTest, connecting_twoice_returns_false)
+{
+  std::unique_ptr<DashboardClient> dashboard_client;
```
urrsk, line 243: Why do you make a new dashboard client object and reset the communication again when that job already has been handle in the test class and its SetUP() function?

tests/test_dashboard_client.cpp
```diff
@@ -219,6 +219,33 @@ TEST_F(DashboardClientTest, connect_non_running_robot)
   EXPECT_LT(elapsed, 2 * comm::TCPSocket::DEFAULT_RECONNECTION_TIME);
 }
 
+TEST_F(DashboardClientTest, non_expected_result_returns_correctly)
+{
+  std::unique_ptr<DashboardClient> dashboard_client;
```
urrsk, line 224: Why do you make a new dashboard client object and reset the communication again when that job already has been handle in the test class and its SetUP() function?

urrsk (changes requested):

tests/test_dashboard_client.cpp
urfeex: Absolutely correct. I started off the test above where we try to connect to a non-existing robot. Hence, I've copied creating a dashboard client. Should be fixed now.

urrsk (approved):

PR #251, opened 2025-01-24
Add more information about acceleration/velocity parametrization in t…

urfeex:
…rajectory examples

The examples were mainly using either time parametrization or implicit acc/vel parametrization using the default values.

This adds more explicit acc/vel parametrized motions to the examples with some explaining comments.

### DIFF
--- a/examples/trajectory_point_interface.cpp
+++ b/examples/trajectory_point_interface.cpp
@@ -58,8 +58,7 @@ void handleRobotProgramState(bool program_running)
 
 void trajDoneCallback(const urcl::control::TrajectoryResult& result)
 {
-  URCL_LOG_INFO("Trajectory done with result %d", result);
-  ;
+  URCL_LOG_INFO("Trajectory done with result %s", urcl::control::trajectoryResultToString(result).c_str());
   g_trajectory_done = true;
 }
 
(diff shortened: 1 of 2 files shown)


examples/trajectory_point_interface.cpp
```diff
@@ -159,17 +161,27 @@ int main(int argc, char* argv[])
     std::vector<urcl::vector6d_t> points{ { -0.203, 0.263, 0.559, 0.68, -1.083, -2.076 },
                                           { -0.203, 0.463, 0.559, 0.68, -1.083, -2.076 } };
     std::vector<double> motion_durations{ 5.0, 5.0 };
+    std::vector<double> velocities{ 2.0, 2.3 };
+    std::vector<double> accelerations{ 2.5, 2.5 };
     std::vector<double> blend_radii{ 0.0, 0.0 };
 
     // Trajectory execution
```
urrsk, line 168:
```suggestion
    // Trajectory execution of the path that goes through the points twice.
```

examples/trajectory_point_interface.cpp
```diff
@@ -159,17 +161,27 @@ int main(int argc, char* argv[])
...
-                                               points.size());
+                                               points.size() * 2);
     for (size_t i = 0; i < points.size(); i++)
     {
       // setting the cartesian parameter makes it interpret the 6d vector as a pose and use movel
       g_my_driver->writeTrajectoryPoint(points[i], true, motion_durations[i], blend_radii[i]);
     }
 
+    // Same motion, but parametrized with acceleration and velocity
+    motion_durations = { 0.0, 0.0 };
```
urrsk, line 178: Maybe we should have some different "writeTrajecotryPoint() function. Like one without the duration, as it is optional to define it. When we do not need to have a zero vector here

urrsk (approved):
See my comments as optional, though it took me a minut to understand why ```points.size() * 2```

PR #254, opened 2025-01-29
Use joint speed for extrapolation rather than differences

urfeex:
With the differences it could happen that the extrapolation was leading to vibrations if commands come in at a lower rate.

With the pure diffs small numerical differences could lead to positive interference with the extrapolation mechanism. This should be avoided by using the joint speeds instead.

That should avoid the issues seen in https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/1258

### DIFF
--- a/resources/external_control.urscript
+++ b/resources/external_control.urscript
@@ -148,8 +148,9 @@ def set_servo_setpoint(q):
 end
 
 def extrapolate():
-  diff = [cmd_servo_q[0] - cmd_servo_q_last[0], cmd_servo_q[1] - cmd_servo_q_last[1], cmd_servo_q[2] - cmd_servo_q_last[2], cmd_servo_q[3] - cmd_servo_q_last[3], cmd_servo_q[4] - cmd_servo_q_last[4], cmd_servo_q[5] - cmd_servo_q_last[5]]
   cmd_servo_q_last = cmd_servo_q
+  local vel = get_actual_joint_speeds()
+  local diff = [vel[0] * steptime, vel[1] * steptime, vel[2] * steptime, vel[3] * steptime, vel[4] * steptime, vel[5] * steptime]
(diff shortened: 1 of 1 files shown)


resources/external_control.urscript
```diff
@@ -148,8 +148,9 @@ def set_servo_setpoint(q):
 end
 
 def extrapolate():
-  diff = [cmd_servo_q[0] - cmd_servo_q_last[0], cmd_servo_q[1] - cmd_servo_q_last[1], cmd_servo_q[2] - cmd_servo_q_last[2], cmd_servo_q[3] - cmd_servo_q_last[3], cmd_servo_q[4] - cmd_servo_q_last[4], cmd_servo_q[5] - cmd_servo_q_last[5]]
   cmd_servo_q_last = cmd_servo_q
+  local vel = get_actual_joint_speeds()
```
urrsk, line 152:
```suggestion
  local vel = get_target_joint_speeds()
```

resources/external_control.urscript
```diff
@@ -148,8 +148,9 @@ def set_servo_setpoint(q):
...
+  local diff = [vel[0] * steptime, vel[1] * steptime, vel[2] * steptime, vel[3] * steptime, vel[4] * steptime, vel[5] * steptime]
   cmd_servo_q = [cmd_servo_q[0] + diff[0], cmd_servo_q[1] + diff[1], cmd_servo_q[2] + diff[2], cmd_servo_q[3] + diff[3], cmd_servo_q[4] + diff[4], cmd_servo_q[5] + diff[5]]
```
urrsk, line 154:
```suggestion
  cmd_servo_q = cmd_servo_q + (vel * steptime)
```

urrsk:
```suggestion
  cmd_servo_q = cmd_servo_q + (get_target_joint_speeds() * steptime)
```

urfeex: We've tested this today on 3.10 where it didn't work, but since we've updated the minimum required version to 3.14 anyway, we can do it that way.

urrsk: The splines code is also using scalar operations

urfeex: I know. Hence, I don't see a problem here.

urrsk (approved):

PR #256, opened 2025-01-29
Improve limit check

urfeex:
This should reduce the limit check getting triggered and also improve the check's output for debugging purposes.


URJala:
Tested on a real robot, and I have not been able to provoke the protective stop message, so seems to work

urrsk (approved):

PR #257, opened 2025-01-31
RTDEClient: pause and stop in destructor only if running

urfeex:
Otherwise we produce an error when destructing an initialized, but non-started client.


urfeex:
I can reproduce the ASAN issue locally with 5.9.4. Taking this back to a draft.

urrsk:
@codecov-ai-reviewer review

SarveshMalladi (approved):

PR #260, opened 2025-02-06
Add codecov/test-results-action

urrsk:


.github/workflows/ci.yml
```diff
@@ -50,7 +50,12 @@ jobs:
       - name: Create folder for test artifacts
         run: mkdir -p test_artifacts
       - name: test
-        run: cd build && ctest --output-on-failure
+        run: cd build && ctest --output-on-failure --output-junit junit.xml
+      - name: Upload test results to Codecov
+        if: ${{ !cancelled() }}
+        uses: codecov/test-results-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
urfeex, line 58:
```suggestion
          token: ${{ secrets.CODECOV_TOKEN }}
          fail_ci_if_error: true
```
I think it would be good it we see when that fails for whatever reason.

urfeex (approved):
Let's give that a go!

PR #265, opened 2025-02-07
Use ExampleRobotWrapper for initialization in all examples

urfeex:
This way the examples are a lot shorter and initialization is the same everywhere, making things easier to maintain.

### DIFF
--- a/src/example_robot_wrapper.cpp
+++ b/src/example_robot_wrapper.cpp
@@ -57,6 +57,14 @@ ExampleRobotWrapper::ExampleRobotWrapper(const std::string& robot_ip, const std:
   {
     startRobotProgram(autostart_program);
   }
+
+  if (headless_mode | !std::empty(autostart_program))
+  {
+    if (!waitForProgramRunning(500))
+    {
+      URCL_LOG_ERROR("Program did not start running. Is the robot in remote control?");
+    }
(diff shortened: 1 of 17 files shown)


src/example_robot_wrapper.cpp
```diff
@@ -66,34 +74,41 @@ ExampleRobotWrapper::~ExampleRobotWrapper()
...
   if (!dashboard_client_->commandPowerOn())
   {
     URCL_LOG_ERROR("Could not send Power on command");
+    return false;
   }
 
   // Release the brakes
   if (!dashboard_client_->commandBrakeRelease())
   {
     URCL_LOG_ERROR("Could not send BrakeRelease command");
```
SarveshMalladi, line 104: Would the program exit the function when encountering an error?

SarveshMalladi (commented):
There is some codecov failures that need to be cleared

src/example_robot_wrapper.cpp
urfeex: That is up to the program to decide. The way I am currently using the wrapper, this function is called in the Constructor, so the only thing we could do is throw an exception from there.

I don't have a clear opinion about throwing an exception from a constructor. Would you think, we should do that?

PR #267, opened 2025-02-11
Use colored log output and timestamps in default log handler

urfeex:
I've had this lying around when implementing things. Maybe that would be useful to have it in the general library.

**Without** this change:
image

**With** this change:
image


urrsk (approved):
I like the colours, it improves the readability. And I love the timestamp!

PR #271, opened 2025-02-19
UrDriver: Send program in headless mode after creating trajectory and script_command servers

urfeex:
Before, we were sending the script code for execution to the robot and then opening the sockets that the script would connect to.

This has been working fine in most cases, but that fails when too much time passes between the two steps. E.g. I experienced that when having debug output switched on, the robot script was not able to connect to the sockets.

From my understanding this has been overlooked and there is no specific reason why the order was like that. Opening the sockets first and then sending the script seems to make more sense.


urrsk (approved):
Make sense

PR #275, opened 2025-02-24
Remove sending an idle command in quintic spline test

urfeex:
I don't know exactly why it was there but it could break the test quite easily since sending an idle command makes the URScript send a cancelled trajectory result sent back.

@urmahp requesting a review from you as you created the `sendIdle` call initially and I don't necessarily understand why.

The fact that the script sends a trajectory result when it shouldn't will be handled separately.

### DIFF
--- a/tests/test_spline_interpolation.cpp
+++ b/tests/test_spline_interpolation.cpp
@@ -141,11 +141,6 @@ protected:
     ASSERT_TRUE(g_my_robot->isHealthy());
   }
 
-  void sendIdle()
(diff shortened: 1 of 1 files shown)


urmahp (approved):
To be honest I cannot remember why I introduced this. My guess would be that it was either introduced for debugging purposes or because of a delay between when the speedscaling is set in the client library and when it is applied on the robot. None of the less since the tests aren't failing I cannot see any reason not to remove it.

PR #281, opened 2025-02-27
instruction_executor: Allow canceling an instruction

urfeex:
This allows cancelling a running motion with the InstructionExecutor. As executions are blocking, they have to be run in another thread to be able to be canceled. See the test for details.

While implementing #280 I thought I would want to cancel the motion after the PSTOP was hit and started implementing this. It wasn't needed for #280, but I think it would be useful, nevertheless.

As this builds ontop of #280, I'll mark this a draft until that one is merged.


src/ur/instruction_executor.cpp
```diff
@@ -110,3 +116,19 @@ bool urcl::InstructionExecutor::moveL(const urcl::Pose& target, const double acc
...
+  {
+    URCL_LOG_INFO("Cancel motion");
+    driver_->writeTrajectoryControlMessage(urcl::control::TrajectoryControlMessage::TRAJECTORY_CANCEL);
+    std::unique_lock<std::mutex> lock(trajectory_result_mutex_);
+    if (trajectory_done_cv_.wait_for(lock, std::chrono::milliseconds(1000)) == std::cv_status::no_timeout)
+    {
+      return trajectory_result_ == urcl::control::TrajectoryResult::TRAJECTORY_RESULT_CANCELED;
+    }
+  }
+  return false;
```
urrsk, line 133: Maybe it makes sense to let  the calling method can know that it could not cancel the trajectory because the mutex was locked.
Will also suggest covering this in the unit test as well

urrsk (approved):

PR #283, opened 2025-03-06
Remove direct primary and secondary stream from UrDriver

urfeex:
And replace them with the PrimaryClient.

### DIFF
--- a/include/ur_client_library/ur/calibration_checker.h
+++ b/include/ur_client_library/ur/calibration_checker.h
@@ -39,7 +39,9 @@ namespace urcl
  * packages. These are then checked against the used kinematics to see if the correct calibration
  * is used.
  */
-class CalibrationChecker : public comm::IConsumer<primary_interface::PrimaryPackage>
+class [[deprecated("Using the standaline CalibrationChecker consumer to compare the kinematics hash is deprecated. "
+                   "Please use the PrimaryClient instead.")]] CalibrationChecker
+  : public comm::IConsumer<primary_interface::PrimaryPackage>
 {
 public:
   /*!
--- a/include/ur_client_library/ur/calibration_checker.h
+++ b/include/ur_client_library/ur/calibration_checker.h
@@ -39,7 +39,7 @@ namespace urcl
  * packages. These are then checked against the used kinematics to see if the correct calibration
  * is used.
  */
-class [[deprecated("Using the standaline CalibrationChecker consumer to compare the kinematics hash is deprecated. "
+class [[deprecated("Using the standalone CalibrationChecker consumer to compare the kinematics hash is deprecated. "
                    "Please use the PrimaryClient instead.")]] CalibrationChecker
   : public comm::IConsumer<primary_interface::PrimaryPackage>
 {
--- a/src/primary/primary_client.cpp
+++ b/src/primary/primary_client.cpp
@@ -67,6 +67,12 @@ void PrimaryClient::start(const size_t max_num_tries, const std::chrono::millise
   pipeline_->run();
 }
 
+void PrimaryClient::stop()
+{
+  pipeline_->stop();
+  stream_.close();
+}
+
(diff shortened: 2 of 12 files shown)


urfeex:
Rebasing this ontop of #284 should resolve the failing integration tests.

include/ur_client_library/ur/calibration_checker.h
```diff
@@ -39,7 +39,9 @@ namespace urcl
  * packages. These are then checked against the used kinematics to see if the correct calibration
  * is used.
  */
-class CalibrationChecker : public comm::IConsumer<primary_interface::PrimaryPackage>
+class [[deprecated("Using the standaline CalibrationChecker consumer to compare the kinematics hash is deprecated. "
```
urmahp, line 42:
```suggestion
class [[deprecated("Using the standalone CalibrationChecker consumer to compare the kinematics hash is deprecated. "
```

src/primary/primary_client.cpp
```diff
@@ -91,5 +97,68 @@ std::deque<ErrorCode> PrimaryClient::getErrorCodes()
...
+    URCL_LOG_ERROR(error_message.c_str());
+    return false;
+  };
+
+  if (send_script_contents("initial attempt"))
+  {
+    return true;
+  }
+
+  if (reconnectStream())
```
urmahp, line 129: What is the purpose of the reconnect stream, when sending scripts to the robot?

src/ur/ur_driver.cpp
```diff
@@ -591,44 +566,11 @@ rtde_interface::RTDEWriter& UrDriver::getRTDEWriter()
...
-
-  if (send_script_contents("initial attempt"))
-  {
-    return true;
-  }
-
-  if (reconnectSecondaryStream())
-  {
-    return send_script_contents("after reconnecting secondary stream");
   }
```
urmahp, line 572: Is there a specific reason that we remove the runtime_error and just return false, if the primary client is not present? Should we add an error message to the LOG, so the user is able to see why it fails.

urmahp (changes requested):

src/primary/primary_client.cpp
urfeex: If the robot switches between local and remote control mode, the connection to the primary interface doesn't work correctly, anymore. This was a workaround for that problem. If we try to send a script to a no-longer connected primary interface, we reconnect it. This should and will be replaced with a proper reconnection on local/remote mode switch, but at this point I decided to keep the workaround.

src/ur/ur_driver.cpp
urfeex: I guess, that's the same question as above?

src/primary/primary_client.cpp
urmahp: Ahh that makes sense.

src/primary/primary_client.cpp
```diff
@@ -91,5 +97,68 @@ std::deque<ErrorCode> PrimaryClient::getErrorCodes()
...
+}
+
+bool PrimaryClient::checkCalibration(const std::string& checksum)
+{
+  std::shared_ptr<primary_interface::KinematicsInfo> kin_info;
+  while (kin_info == nullptr)
+  {
+    kin_info = consumer_->getKinematicsInfo();
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+  }
```
urmahp, line 157:
```suggestion
  std::shared_ptr<primary_interface::KinematicsInfo> kin_info = consumer_->getKinematicsInfo();
  while (kin_info == nullptr)
  {
    std::this_thread::sleep_for(std::chrono::seconds(1));
    kin_info = consumer_->getKinematicsInfo();
  }
```
Small optimization so we don't have to wait 1 seconds if the kinematics info is present to begin with.

src/ur/ur_driver.cpp
urfeex: Now I understand. I will add the exception, again.

PR #289, opened 2025-03-13
Primary client power on

urfeex:
This allows power-cycling the robot remotely without using the dashboard server.

### DIFF
--- a/src/primary/primary_client.cpp
+++ b/src/primary/primary_client.cpp
@@ -31,6 +31,7 @@
 #include <ur_client_library/primary/primary_client.h>
 #include <ur_client_library/primary/robot_message.h>
 #include <ur_client_library/primary/robot_state.h>
+#include "ur_client_library/exceptions.h"
 
 namespace urcl
 {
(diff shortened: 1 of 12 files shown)


src/primary/primary_client.cpp
```diff
@@ -160,5 +160,60 @@ bool PrimaryClient::checkCalibration(const std::string& checksum)
...
+    return false;
+  }
+  if (validate)
+  {
+    return waitFor([this]() { return getRobotMode() == RobotMode::RUNNING; }, timeout);
+  }
+  return true;
+}
+
+bool PrimaryClient::waitFor(std::function<bool()> condition, const std::chrono::milliseconds timeout,
```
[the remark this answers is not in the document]
urfeex: yes, that was a mistake apparently.

src/primary/primary_client.cpp
```diff
@@ -31,6 +31,7 @@
 #include <ur_client_library/primary/primary_client.h>
 #include <ur_client_library/primary/robot_message.h>
 #include <ur_client_library/primary/robot_state.h>
+#include "ur_client_library/exceptions.h"
```
urmahp, line 35:
```suggestion
#include <ur_client_library/helpers.h>
```
In my opinion we should be explicit and inlcude the helpers.h file here for the wait_for function. I am however a bit surprised that it doesn't complain when compiling.

urmahp (changes requested):
Almost there, just one small suggestion.

src/primary/primary_client.cpp
urfeex: It probably gets pulled in recursively. But I agree.

urmahp (approved):

PR #292, opened 2025-03-17
PrimaryClient: Add methods to unlock protective stop and stop the program

urfeex:
Add functionalities to stop the running program and to unlock a protective stop through the primary client.

### DIFF
--- a/include/ur_client_library/primary/primary_client.h
+++ b/include/ur_client_library/primary/primary_client.h
@@ -188,13 +188,15 @@ public:
    *
    * The robot's protective_stop state will be updated in the background. This will always show the latest received
    * state independent of the time that has passed since receiving it.
+   *
+   * \throws UrException when no robot mode data has been received, yet.
    */
   bool isRobotProtectiveStopped()
   {
     std::shared_ptr<RobotModeData> robot_mode_data = consumer_->getRobotModeData();
     if (robot_mode_data == nullptr)
     {
-      return false;
+      throw UrException("Robot mode data is a nullptr. Probably it hasn't been received, yet.");
     }
     return robot_mode_data->is_protective_stopped_;
   }
--- a/include/ur_client_library/primary/primary_client.h
+++ b/include/ur_client_library/primary/primary_client.h
@@ -129,6 +129,39 @@ public:
                            const std::chrono::milliseconds timeout = std::chrono::seconds(30));
 
   /*!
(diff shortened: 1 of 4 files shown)


urfeex:
I moved the test modification to an own PR #293. Let's merge that first.

src/primary/primary_client.cpp
```diff
@@ -212,12 +212,75 @@ void PrimaryClient::commandBrakeRelease(const bool validate, const std::chrono::
...
+      waitFor([this]() { return consumer_->getRobotModeData()->is_protective_stopped_ == false; }, timeout);
+    }
+    catch (const TimeoutException&)
+    {
+      throw TimeoutException("Robot did not unlock the protective stop within the given timeout", timeout);
+    }
+  }
+}
+
+void PrimaryClient::commandStop(const bool already_stopped_ok, const bool validate,
```
urrsk, line 241: I am not sure that it is obvious what the arguments mean. Special the 'validate', another name suggestion could be "blocking"

In addition, the functionality of `already_stopped_ok` might be challeging with a program that finish within 100ms inteval that the primary interface is updating the with. An example of this could be a small program setting an output. 

We could maybe consider using enum for "modes"

urfeex: > I am not sure that it is obvious what the arguments mean. Special the 'validate', another name suggestion could be "blocking"

I would be fine with that but then I would update that in all methods.

> In addition, the functionality of already_stopped_ok might be challeging with a program that finish within 100ms inteval that the primary interface is updating the with. An example of this could be a small program setting an output.

Isn't that exactly an example of why we need such a flag?

>  We could maybe consider using enum for "modes"

Which modes are you referring to?

include/ur_client_library/primary/primary_client.h
```diff
@@ -144,6 +171,34 @@ class PrimaryClient
...
+   *
+   * The robot's protective_stop state will be updated in the background. This will always show the latest received
+   * state independent of the time that has passed since receiving it.
+   */
+  bool isRobotProtectiveStopped()
+  {
+    std::shared_ptr<RobotModeData> robot_mode_data = consumer_->getRobotModeData();
+    if (robot_mode_data == nullptr)
+    {
+      return false;
```
urrsk, line 197: I kind of more expecting an exception to be thrown here!

urrsk (approved):

PR #294, opened 2025-03-20
Make start_ursim.sh support polyscopex

urfeex:
This adds support for starting a PolyScopeX-based URsim. The default version will still be the latest PolyScope 5 version.

At this moment, this does not install the External Control URCapX as we don't have a download of the built package available.


scripts/start_ursim.sh
```diff
@@ -136,6 +141,43 @@ validate_ursim_version()
...
+  do
+    sleep 1
+    #echo "still waiting for UrService..."
+    echo -ne "."
+    status_code=$(eval "$curl_cmd")
+  done
+
+  echo ""; echo "UrService is up"
+
+  # TODO: Once we have a downloadable URCapX, we can use the following code to install it
```
domire8, line 172: Hi @urfeex sorry for peeking, what is the expected release here?

urfeex: I cannot give a date, but we are trying to finalize things at the moment. The version from [UniversalRobots/Universal_Robots_ExternalControl_URCapX](https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCapX) should work mostly fine at the moment, with https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCapX/pull/7 being merged soon, I guess. 

This has to be built manually at the moment, which is why I placed the comment in the script for now. We plan to have a CI and a released .urcapx in the repo soon.

domire8: Ok sure I understand. Thanks for the reply! We might try this out in the next couple of weeks :+1:

scripts/start_ursim.sh
```diff
@@ -34,6 +34,13 @@ IP_ADDRESS="192.168.56.101"
 PORT_FORWARDING="-p 30001-30004:30001-30004 -p 29999:29999"
```
urmahp, line 34: Since dashboard server is not present on PolyscopeX, the port 29999 should not be forwarded. Maybe create another port forwarding for PolyscopeX

scripts/start_ursim.sh
```diff
@@ -136,6 +141,43 @@ validate_ursim_version()
...
+{
+
+  echo -ne "Starting URSim. Waiting for UrService to be up..."
+  curl_cmd="curl --retry-connrefused -f --write-out %{http_code} --silent --output /dev/null $IP_ADDRESS/universal-robots/urservice/api/v1/urcaps"
+  status_code=$(eval "$curl_cmd")
+
+  until [ "$status_code" == "200" ]
+  do
+    sleep 1
+    #echo "still waiting for UrService..."
```
urmahp, line 165:
```suggestion
```
Remove

scripts/start_ursim.sh
```diff
@@ -175,44 +217,70 @@ done
...
+    -v ${URCAP_STORAGE}:/urcaps \
+    -v ${PROGRAM_STORAGE}:/ursim/programs \
+    -e ROBOT_MODEL=${ROBOT_MODEL} \
+    $PORT_FORWARDING \
+    --name $CONTAINER_NAME \
+    universalrobots/ursim_${ROBOT_SERIES}:$URSIM_VERSION"
+fi
 
-#echo $docker_cmd
+#echo "$docker_cmd"
```
urmahp, line 283: Is there a specific reason to have this comment?

scripts/start_ursim.sh
domire8: Tried it out and works like a charm so far :+1:

scripts/start_ursim.sh
```diff
@@ -175,44 +217,70 @@ done
 validate_model
 validate_ursim_version
 
+if [ "$URSIM_VERSION" != "latest" ]; then
```
urmahp, line 220: Could we add this check to validate_ursim_version, below is a suggestion of how it could be done. 
```
validate_ursim_version()
{
  local IMAGE_URSIM_VERSION
  # Inspect the image's URSim version if the image is locally available. This is especially
  # important when we use the "latest" tag, as we don't know the version hiding behind this and it
  # could be potentially older.
  IMAGE_URSIM_VERSION=$(docker image inspect universalrobots/ursim_"${ROBOT_SERIES}":"$URSIM_VERSION" 2>/dev/null | grep -Po '"build_version": "URSim Version: \K[^"]*') || true
  if [ -z "$IMAGE_URSIM_VERSION" ]; then
    IMAGE_URSIM_VERSION="$URSIM_VERSION"
  fi
  [ "$IMAGE_URSIM_VERSION" == "latest" ] && return 0
  local MIN_CB3="3.14.3"
  local MIN_E_SERIES="5.9.4"
  local MIN_UR20="5.14.0"
  local MIN_UR30="5.15.0"

  local MIN_VERSION="0.0"

if [ "$URSIM_VERSION" != "latest" ]; then
  verlte "10.0.0" "$URSIM_VERSION" && ROBOT_SERIES=polyscopex
fi


  case $ROBOT_SERIES in
    cb3)
      verlte "4.0.0" "$IMAGE_URSIM_VERSION" && echo "$IMAGE_URSIM_VERSION is no valid CB3 version!" && exit
      verlte "$MIN_CB3" "$IMAGE_URSIM_VERSION" && return 0
      MIN_VERSION=$MIN_CB3
      ;;
    e-series)
      if [[ $ROBOT_MODEL == "UR20" ]]; then
          verlte "$MIN_UR20" "$IMAGE_URSIM_VERSION" && return 0
          MIN_VERSION=$MIN_UR20
      elif [[ $ROBOT_MODEL == "UR30" ]]; then
          verlte "$MIN_UR30" "$IMAGE_URSIM_VERSION" && return 0
          MIN_VERSION=$MIN_UR30
      else
          verlte "$MIN_E_SERIES" "$URSIM_VERSION" && return 0
          MIN_VERSION=$MIN_E_SERIES
      fi
      ;;
 polyscopeX)
 if [ $ROBOT_MODEL in
    ur3|ur5|ur10) ]; then
      echo "$ROBOT_MODEL is no valid PolyscopeX version!" && exit
  else
    return 0
   fi
;;
  esac

  echo "Illegal version given. For $ROBOT_SERIES $ROBOT_MODEL the software version must be greater or equal to $MIN_VERSION. Given version: $IMAGE_URSIM_VERSION."
  exit
}

```

urmahp, line 220: I think a robot_series argument should be added to the script, such that the latest tag works with PolyscopeX as well.

scripts/start_ursim.sh
```diff
@@ -175,44 +217,70 @@ done
...
+  verlte "10.0.0" "$URSIM_VERSION" && ROBOT_SERIES=polyscopex
+fi
+URCAP_STORAGE="${PERSISTENT_BASE}/${ROBOT_SERIES}/urcaps"
+PROGRAM_STORAGE="${PERSISTENT_BASE}/${ROBOT_SERIES}/${ROBOT_MODEL}/programs"
+
+DOCKER_ARGS=""
+
+if [ "$ROBOT_SERIES" == "polyscopex" ]; then
```
urmahp, line 228: This could also be added to the ursim validation function, it is also validating the ursim version

urmahp (changes requested):

scripts/start_ursim.sh
urfeex: I use it quite often when I debug things. But we can also remove it. I guess, it ended up in here as I forgot to remove it at some point.

scripts/start_ursim.sh
urfeex: I am not entirely sure about that. This would also require to have a "latest" working for CB3, which would require a manual mapping for that. Also, the PolyScopeX image doesn't have a latest tag. This is not a big deal as we have the mapping for that, anyway. I would vote we go without that for now.

urfeex: For CB3 and e-series we infer the series from the model, hence the model has to be parsed before validating the URSim version. After parsing the model, it contains the representation used for URSim, where a UR5 and a UR5e is a "UR5" in the respective URSim version. Hence, when checking the robot model in the `validate_ursim_version()` function,  the `e` will be removed, anyway.

It might be worth refactoring this part, however, also considering future models.

urfeex:
@urmahp I've done some major refactoring with more validation and tests of that validation. This should allow us to react on future changes in a more robust way that we did in the past.

scripts/start_ursim.sh
```diff
@@ -31,8 +31,17 @@
...
+PORT_FORWARDING_WITHOUT_DASHBOARD="-p 30001-30004:30001-30004"
 CONTAINER_NAME="ursim"
+TEST_RUN=false
+
+# TODO: Add support for more URSim PolyScopeX versions once released
+# The PolyScopeX URSim containers follow the SDK versioning scheme. This maps those to marketing
+# versions
+#declare -Ag POLYSCOPE_X_MAP=( ["10.7.0"]="0.12.159"
+                             #["10.8.0"]="not.there.yet")
+declare -Ag POLYSCOPE_X_MAP=( ["10.7.0"]="0.12.159" )
```
EbbeFuglsang, line 44:
```suggestion
declare -Ag POLYSCOPE_X_MAP=( 
	["10.7.0"]="0.12.159" 
	["10.8.0"]="0.13.106" 
)
```
10.8.0 is available here https://hub.docker.com/r/universalrobots/ursim_polyscopex/tags

EbbeFuglsang: I think you are right. So maybe it is just a 10.8.0.0😉

urfeex: Would you be OK with leaving this out of this PR for now?

EbbeFuglsang: Yes, that is fine. Maybe we can consider how we are tagging the PolyScope X ursim

urmahp (approved):
Looks very good. Very nice with the added test.

PR #296, opened 2025-03-20
Reduce usage of dashboard client in tests and examples

urfeex:
As we now have more capabilities in the primary interface, we don't need to rely on the dashboard client to initialize the robot in the example robot wrapper.

### DIFF
--- a/src/example_robot_wrapper.cpp
+++ b/src/example_robot_wrapper.cpp
@@ -32,6 +32,7 @@
 #include <iostream>
 #include "ur_client_library/exceptions.h"
 #include "ur_client_library/log.h"
+#include "ur_client_library/ur/version_information.h"
 
 namespace urcl
 {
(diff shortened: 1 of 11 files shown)


src/example_robot_wrapper.cpp
```diff
@@ -94,20 +102,28 @@ ExampleRobotWrapper::~ExampleRobotWrapper()
...
       std::this_thread::sleep_for(std::chrono::seconds(5));
-      if (!dashboard_client_->commandUnlockProtectiveStop())
+      try
       {
-        URCL_LOG_ERROR("Could not unlock protective stop");
+        primary_client_->commandUnlockProtectiveStop();
+      }
+      catch (const TimeoutException&)
+      {
+        URCL_LOG_ERROR("Robot did not unlock the protective stop within the given timeout");
```
urmahp, line 126: The user doesn't specify a timeout, so I prefer a slightly modification of the old log message. However, if you prefer the new log message I am also totally fine with keeping the new log message. 
```suggestion
        URCL_LOG_ERROR("Robot could not unlock the protective stop");
```

urmahp (approved):
Overall looks good and I am fine with merging as it is. I only have a mall suggestion for a logging message.

PR #307, opened 2025-04-04
Fix naming issues

### LINKED ISSUE #297: Variable names with underscore
https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/b42da6d59ad2e6259809a547aea9c3de4218179b/include/ur_client_library/ur/ur_driver.h#L128

The configuration struct for the UrDriver has two members that have a name with an underscore at the end:
- `rtde_initialization_attempts_`
- `rtde_initialization_timeout_`

If that's not desired, maybe worth fixing

urfeex:
This fixes #297.

Merging this PR will be breaking public API and will therefore require a major bump. It has still to be decided how big of a change we want to do with that.

We have been a bit inconsistent in the past when it comes to naming public members. 

While the changes in 71bab65250477a09a79bf112d860e146ef517a57 and 42cf1473998b0747e7504aa0bf7bdcda10f2067b would probably not impact many people, the changes from bef94bbe7cc2d49d1c3d9167bbc164d7544504c6 would be more impactful.

Potentially there would be many more changes in the message structures for the primary and RTDE interfaces which I have disabled for now.

Please consider this PR up for a debate. We want to fix #297 and avoid that from happening in future, it's just the question of how far we break the API in order to get things aligned.

### DIFF
--- a/include/ur_client_library/example_robot_wrapper.h
+++ b/include/ur_client_library/example_robot_wrapper.h
@@ -169,17 +169,30 @@ public:
 
   bool isHealthy() const;
 
-  //! Dashboard client to interact with the robot
-  std::shared_ptr<urcl::DashboardClient> dashboard_client;
-
-  //! primary client to interact with the robot
-  std::shared_ptr<urcl::primary_interface::PrimaryClient> primary_client;
-
-  //! UR driver to interact with the robot
-  std::shared_ptr<urcl::UrDriver> ur_driver;
+  std::shared_ptr<DashboardClient> getDashboardClient() const
+  {
+    return dashboard_client_;
+  };
+  std::shared_ptr<primary_interface::PrimaryClient> getPrimaryClient() const
+  {
+    return primary_client_;
+  };
+  std::shared_ptr<UrDriver> getUrDriver() const
+  {
+    return ur_driver_;
+  };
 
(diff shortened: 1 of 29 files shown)


include/ur_client_library/example_robot_wrapper.h
```diff
@@ -170,13 +170,13 @@ class ExampleRobotWrapper
   bool isHealthy() const;
 
   //! Dashboard client to interact with the robot
-  std::shared_ptr<urcl::DashboardClient> dashboard_client_;
+  std::shared_ptr<urcl::DashboardClient> dashboard_client;
```
urmahp, line 173: Is there a reason that we have these variables as public, instead of accessing them through a get function and then making the variables private?

urmahp (commented):
I think we should consider whether we want to make the variables private and then access them through get functions instead, but other than that I think it looks fine.

include/ur_client_library/example_robot_wrapper.h
urfeex: I think, the main reason is that I just made them public at some point to have them accessible in the tests. Getters seems like the cleaner solution. I'll update that.

urmahp (approved):

PR #318, opened 2025-05-06
Allow an empty input recipe to the RTDE client

urfeex:
This PR aims at providing the option to use an RTDE client READ-ONLY by providing an empty input recipe.

At this stage it is only an attempt to play around with that possibility, implementation details may change significantly.

### DIFF
--- a/include/ur_client_library/rtde/rtde_writer.h
+++ b/include/ur_client_library/rtde/rtde_writer.h
@@ -171,7 +171,7 @@ private:
   uint8_t pinToMask(uint8_t pin);
   comm::URStream<RTDEPackage>* stream_;
   std::vector<std::string> recipe_;
-  uint8_t recipe_id_ = 0;
+  uint8_t recipe_id_;
   moodycamel::BlockingReaderWriterQueue<std::unique_ptr<DataPackage>> queue_;
   std::thread writer_thread_;
   bool running_;
--- a/src/rtde/rtde_writer.cpp
+++ b/src/rtde/rtde_writer.cpp
@@ -33,7 +33,7 @@ namespace urcl
 namespace rtde_interface
 {
 RTDEWriter::RTDEWriter(comm::URStream<RTDEPackage>* stream, const std::vector<std::string>& recipe)
-  : stream_(stream), recipe_(recipe), queue_{ 32 }, running_(false), package_(recipe_)
+  : stream_(stream), recipe_(recipe), recipe_id_(0), queue_{ 32 }, running_(false), package_(recipe_)
 {
 }
 


include/ur_client_library/rtde/rtde_writer.h
```diff
@@ -162,7 +171,7 @@ class RTDEWriter
   uint8_t pinToMask(uint8_t pin);
   comm::URStream<RTDEPackage>* stream_;
   std::vector<std::string> recipe_;
-  uint8_t recipe_id_;
+  uint8_t recipe_id_ = 0;
```
urmahp: Could we initialize this in the constructor instead?

tests/test_rtde_client.cpp
```diff
@@ -413,6 +416,48 @@ TEST_F(RTDEClientTest, check_unknown_rtde_output_variable)
   EXPECT_THROW(client_->init(), UrException);
 }
 
+TEST_F(RTDEClientTest, empty_input_recipe)
+{
+  std::vector<std::string> empty_input_recipe = {};
```
urmahp, line 421: I think we should add a test with an empty input recipe file as well. The behavior is not exactly the same and this doesn't cover that.

urfeex: In fact, [we have that test, already](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/2bb37c2af88be94a0ddfaebbfd8f26b2a3fd5b1b/tests/test_rtde_client.cpp#L131-L143). 

The question would be whether or not we want to change the behaviour of that. Currently

* Passing an empty dict or and empty filename `""` to the RTDEClient's initialization is valid and will result in a read-only client
* Passing a non-existing filename raises an exception
* Passing a valid filename with empty content raises an exception (that's the use case we are discussing here, right?)

In my opinion that makes sense, since if you want to pass an input recipe to the client, then you probably also want to set an input recipe. However, allowing empty input recipe files would allow making an existing application use a read-only client using only the runtime configuration. (Strictly, that would also be possible if the recipe file's path is a runtime-argument of the application and allows passing an empty string). I am favoring the way it currently is, but I am not absolutely against changing it. What do you think @urmahp?

urmahp: I agree we should keep the changes as is. 
I was more thinking of testing that we can read from a stream when initialized with "", but I guess that is sort of tested with https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/2bb37c2af88be94a0ddfaebbfd8f26b2a3fd5b1b/tests/test_rtde_client.cpp#L114-L129, at least we test that it doesn't throw an exception.

PR #323, opened 2025-05-16
Added command to set friction compensation for the torque command

urmahp:


include/ur_client_library/control/script_command_interface.h
```diff
@@ -147,6 +147,17 @@ class ScriptCommandInterface : public ReverseInterface
...
+  /*!
+   * \brief Set friction compensation for the torque_command. If true the torque command will compensate for friction,
+   * if false it will not.
+   *
+   * \param friction_compesation Will set a friction_compensation variable in urscript, which will be used when calling
+   * torque_command
+   *
+   * \returns True, if the write was performed successfully, false otherwise.
+   */
+  bool setFrictionCompensation(const bool friction_compensation);
```
urfeex, line 159: I've seen many discussions around toggling  a feature on and off, in my opinion having a clear parameter name that leaves no ambiguities what a `true` value means is helpful in these situations.

```suggestion
   * \param friction_compensation_enabled Will set a friction_compensation variable in urscript, which will be used when calling
   * torque_command
   *
   * \returns True, if the write was performed successfully, false otherwise.
   */
  bool setFrictionCompensation(const bool friction_compensation_enabled);
```

include/ur_client_library/ur/ur_driver.h
```diff
@@ -714,6 +714,17 @@ class UrDriver
...
+  /*!
+   * \brief Set friction compensation for the torque_command. If true the torque command will compensate for friction,
+   * if false it will not.
+   *
+   * \param friction_compesation Will set a friction_compensation variable in urscript, which will be used when calling
+   * torque_command
+   *
+   * \returns True, if the write was performed successfully, false otherwise.
+   */
+  bool setFrictionCompensation(const bool friction_compensation);
```
urfeex, line 726:
```suggestion
   * \param friction_compensation_enabled Will set a friction_compensation variable in urscript, which will be used when calling
   * torque_command
   *
   * \returns True, if the write was performed successfully, false otherwise.
   */
  bool setFrictionCompensation(const bool friction_compensation_enabled);
```

src/control/script_command_interface.cpp
```diff
@@ -222,6 +222,29 @@ bool ScriptCommandInterface::endToolContact()
   return server_.write(client_fd_, buffer, sizeof(buffer), written);
 }
 
+bool ScriptCommandInterface::setFrictionCompensation(const bool friction_compensation)
```
urfeex, line 225:
```suggestion
bool ScriptCommandInterface::setFrictionCompensation(const bool friction_compensation_enabled)
```

src/control/script_command_interface.cpp
```diff
@@ -222,6 +222,29 @@ bool ScriptCommandInterface::endToolContact()
...
+{
+  const int message_length = 2;
+  uint8_t buffer[sizeof(int32_t) * MAX_MESSAGE_LENGTH];
+  uint8_t* b_pos = buffer;
+
+  int32_t val = htobe32(toUnderlying(ScriptCommand::SET_FRICTION_COMPENSATION));
+  b_pos += append(b_pos, val);
+
+  val = htobe32(friction_compensation);
```
urfeex, line 234:
```suggestion
  val = htobe32(friction_compensation_enabled);
```

src/ur/ur_driver.cpp
```diff
@@ -527,6 +527,19 @@ bool UrDriver::endToolContact()
   }
 }
 
+bool UrDriver::setFrictionCompensation(const bool friction_compensation)
+{
+  if (script_command_interface_->clientConnected())
+  {
+    return script_command_interface_->setFrictionCompensation(friction_compensation);
```
urfeex, line 534:
```suggestion
bool UrDriver::setFrictionCompensation(const bool friction_compensation_enabled)
{
  if (script_command_interface_->clientConnected())
  {
    return script_command_interface_->setFrictionCompensation(friction_compensation_enabled);
```

src/ur/ur_driver.cpp
```diff
@@ -527,6 +527,19 @@ bool UrDriver::endToolContact()
   }
 }
 
+bool UrDriver::setFrictionCompensation(const bool friction_compensation)
```
urfeex, line 530: It might be great to start testing the script command functions from the ur_driver object.

urfeex (changes requested):

src/ur/ur_driver.cpp
urmahp: Yes that is a good idea. But we should probably not start doing it on a beta branch, but start by adding the first test to the master branch, which would then test some of the other commands.

doc/architecture/script_command_interface.rst
```diff
@@ -121,6 +123,15 @@ The robot reads from the "script_command_socket" expecting a 32 bit integer repr
    1      No specific meaning / values ignored
    =====  =====
 
+.. table:: With endToolContact command
```
urfeex, line 126:
```suggestion
.. table:: With setFrictionCompensation command
```

urfeex (approved):
With my suggestion that looks fine.

PR #325, opened 2025-05-19
Add torque control to urscript

domire8:
The PR adds `MODE_TORQUE` to the supported control modes and extends the external control urscript with a torque thread that applies the latest `cmd_tau` as torque command.

`torque_command` also has an optional parameter `friction_comp` that might be worth parametrizing too, but I'm not sure what effect that had in the first place so I went hardcoded for an MVP.

Let me know what you think about this and what might still be missing.

### DIFF
--- a/resources/external_control.urscript
+++ b/resources/external_control.urscript
@@ -71,7 +71,7 @@ JOINT_IGNORE_SPEED = 20.0
 global violation_popup_counter = 0
 global cmd_servo_state = SERVO_UNINITIALIZED
 global cmd_servo_qd = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
-global cmd_tau = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+global cmd_torque = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 global cmd_servo_q = get_joint_positions()
 global cmd_servo_q_last = cmd_servo_q
 global cmd_twist = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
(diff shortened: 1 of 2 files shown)


resources/external_control.urscript
```diff
@@ -225,6 +227,22 @@ thread speedThread():
...
+
+thread torqueThread():
+  textmsg("ExternalControl: Starting torque thread")
+  while control_mode == MODE_TORQUE:
+    tau = cmd_tau
+    torque_command(tau, friction_comp=True)
+  end
+  textmsg("ExternalControl: torque thread ended")
+  stopj(STOPJ_ACCELERATION)
+end
```
urfeex, line 244: We have added an interface to dynamically switch friction compensation on and off in #323 
```suggestion
thread torqueThread():
  textmsg("ExternalControl: Starting torque thread")
  while control_mode == MODE_TORQUE:
    torque = cmd_torque
    torque_command(torque, friction_comp=friction_compensation_enabled)
  end
  textmsg("ExternalControl: torque thread ended")
  stopj(STOPJ_ACCELERATION)
end
```

resources/external_control.urscript
```diff
@@ -766,6 +786,9 @@ while control_mode > MODE_STOPPED:
     elif control_mode == MODE_SPEEDJ:
       qd = [params_mult[2] / MULT_jointstate, params_mult[3] / MULT_jointstate, params_mult[4] / MULT_jointstate, params_mult[5] / MULT_jointstate, params_mult[6] / MULT_jointstate, params_mult[7] / MULT_jointstate]
       set_speed(qd)
+    elif control_mode == MODE_TORQUE:
+      tau = [params_mult[2]/ MULT_jointstate, params_mult[3]/ MULT_jointstate, params_mult[4]/ MULT_jointstate, params_mult[5]/ MULT_jointstate, params_mult[6]/ MULT_jointstate, params_mult[7]/ MULT_jointstate]
+      set_torque(tau)
```
urfeex, line 791:
```suggestion
      torque = [params_mult[2]/ MULT_jointstate, params_mult[3]/ MULT_jointstate, params_mult[4]/ MULT_jointstate, params_mult[5]/ MULT_jointstate, params_mult[6]/ MULT_jointstate, params_mult[7]/ MULT_jointstate]
      set_torque(torque)
```

urfeex (changes requested):
First of all, thank you very much for contributing this!

I mainly have two points at this stage:

- In my opinion it would be better to use `torque` instead of `tau` for readability. I added my perspective inside the first occurrence. Would you agree with that?
- Second, we've added an interface to set the friction compensation parameter on and off in #323, so we can use that in the `torque_command` call directly.

urfeex (approved):
Thanks again for implementing and contributing this!

PR #326, opened 2025-05-19
Add an example using the script_command_interface

urfeex:
This adds an example for using the script_command interface. It basically came out of testing #323 manually.

### DIFF
--- a/examples/script_command_interface.cpp
+++ b/examples/script_command_interface.cpp
@@ -1,0 +1,142 @@
+// -- BEGIN LICENSE BLOCK ----------------------------------------------
+// Copyright 2025 Universal Robots A/S
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright
+//      notice, this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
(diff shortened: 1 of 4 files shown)


examples/script_command_interface.cpp
```diff
@@ -0,0 +1,142 @@
...
+      std::this_thread::sleep_for(timeout);
+      URCL_LOG_INFO(log_output.c_str());
+      func();
+    }
+  };
+
+  // Keep running all commands in a loop until g_running is set to false
+  while (g_running)
+  {
+    run_cmd("Setting tool voltage to 24V",
```
urmahp, line 64: We should call start and stop force mode as well

examples/script_command_interface.cpp
```diff
@@ -0,0 +1,142 @@
...
+const std::string INPUT_RECIPE = "examples/resources/rtde_input_recipe.txt";
+
+std::unique_ptr<ExampleRobotWrapper> g_my_robot;
+bool g_HEADLESS = true;
+bool g_running = false;
+
+void sendScriptCommands()
+{
+  auto run_cmd = [](const std::string& log_output, std::function<void()> func) {
+    const std::chrono::seconds timeout(3);
```
urmahp, line 52: Is there a reason to sleep for 3 seconds between each call? Is it to not call each command immediately after each other?

examples/script_command_interface.cpp
urfeex: As force mode needs quite a lot of arguments and might complain about singularities if the robot is in the wrong pose, I would rather skip it here. It's got its own example. But I can also add freedrive mode.

urmahp (approved):

Issue #340, opened 2025-06-03
The endian.h header from 3rdparty needs to be installed


acmorrow:
### Affected Client Library version(s)

2.0..0

### What combination of platform is the Client Library running on.

Windows, Mac

### How did you install the Client Library

Build the library from source

### Which robot platform is the library connected to.

URSim in docker

### Robot SW / URSim version(s)

N/A

### How are you using the Client Library

Others

### Issue details

# Summary
When experimenting with building the client library on macOS, I found that it doesn't currently handle the lack of `endian.h` on that platform. There is a workaround for MSVC in the top-level CMakeLists.txt, where it says:
```
if(MSVC)
  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/endian)
```
As an experiment, I added similar logic for `if (APPLE`.

This was sufficient to get the client library to build, but it is not sufficient for consumers of the client library since:
1) The endian header is included from public includes:
```
build/install/include/ur_client_library/comm/package_serializer.h:#include <endian.h>
build/install/include/ur_client_library/comm/bin_parser.h:#include <endian.h>
build/install/include/ur_client_library/comm/bin_parser.h:  // Decode from network encoding (big endian) to host encoding
build/install/include/ur_client_library/rtde/package_header.h:#include <endian.h>
build/install/include/ur_client_library/control/reverse_interface.h:#include <endian.h>
build/install/include/ur_client_library/primary/package_header.h:#include <endian.h>
```

2) The need for the additional include path isn't propagated to consumers of the library, e.g. there isn't a stanza like this for it that attaches the `endian.h` headers path where the header should be found at build time and from the installation:

```
target_include_directories( urcl PUBLIC
  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
  $<INSTALL_INTERFACE:include>
)
```

3) The `3rdparty/endian.h` header doesn't have any installation rules that would get it populated into the installation tree.

## Steps to Reproduce
I found this by messing around on macOS with some local changes, but my guess would be that this would be an immediate issue on Windows too. I think if you build the client library, install it, and write a small consuming program which directly or indirectly includes any of the above headers it will fail because it can't find the `endian.h` header.

### Relevant log output

```shell

```

### Accept Public visibility

- [x] I agree to make this context public

urfeex:
Thank you, that argument seems reasonable.

PR #341, opened 2025-06-03
Minimal support for building on macOS

acmorrow:
I'd like to be able to do iterative development against the UR client library on macOS since that's my local platform. I don't need to actually run anything there, necessarily, though it would, ultimately, be nice to have the simulator working there too. This is enough to get the client library to build from source on macOS. Note though that it is incomplete due to #340 - the installed library cannot yet be consumed since the `endian.h` header doesn't get installed.


urfeex (approved):
Seems fine with me. If that helps getting this to build on MacOS, all the better.

PR #342, opened 2025-06-05
Added commands to set pd controller gains and  maximum joint torques

urmahp:


include/ur_client_library/control/script_command_interface.h
```diff
@@ -158,6 +158,30 @@ class ScriptCommandInterface : public ReverseInterface
    */
   bool setFrictionCompensation(const bool friction_compensation_enabled);
 
+  /*!
+   * \brief Set gains for the PD controller running in the external control script. The PD controller computes joint
+   * torques based on either tcp poses or joint poses and applies the torques to the robot using the torque_command
+   * function. The gains can be used to change the response of the controller. Be aware that changing the controller
+   * response can make it unstable.
```
urfeex, line 165:
```suggestion
   * response can make it unstable.
   * The PD controller can be used without explicitly defining those gains, as it contains a set of default gains for each robot model.
```

include/ur_client_library/ur/ur_driver.h
```diff
@@ -725,6 +725,30 @@ class UrDriver
    */
   bool setFrictionCompensation(const bool friction_compensation_enabled);
 
+  /*!
+   * \brief Set gains for the PD controller running in the external control script. The PD controller computes joint
+   * torques based on either tcp poses or joint poses and applies the torques to the robot using the torque_command
+   * function. The gains can be used to change the response of the controller. Be aware that changing the controller
+   * response can make it unstable.
```
urfeex, line 732: Same as above

tests/test_script_command_interface.cpp
```diff
@@ -421,6 +421,76 @@ TEST_F(ScriptCommandInterfaceTest, test_set_friction_compensation)
...
+  urcl::vector6d_t kp = { 220.2, 220.2, 300.0, 10.32, 10.32, 10.32 };
+  urcl::vector6d_t kd = { 29.68, 29.68, 35.0, 6.4, 6.4, 6.4 };
+  script_command_interface_->setPDControllerGains(&kp, &kd);
+
+  int32_t command;
+  std::vector<int32_t> message;
+  client_->readMessage(command, message);
+
+  // 8 is set PD controller gains
+  int32_t expected_command = 8;
```
urfeex, line 438: We could think about moving the enum to a public namespace so that can be reused here. or is there a specific reason why [`enum class ScriptCommand`](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/165a7c1fca44de845257b8a006ce891388dde559/include/ur_client_library/control/script_command_interface.h#L212) is a private member of `ScriptCommandInterface`?

That's probably not for this PR, however.

urfeex (approved):
I only have minor documentation suggestions. Feel free to accept or ignore them.

tests/test_script_command_interface.cpp
urmahp: No I think its because it is only intended to be used when sending commands to the external control script. But would make sense to make them available for the test cases. But yes should probably be done in separate PR.

PR #343, opened 2025-06-12
Add ScriptReader for script template parsing

urfeex:
This PR adds a script parser that extends the templating capabilities when loading script code.

In particular that allows
- Variable substitution (exists already, but this was refactored)
- Conditionals
- Includes

With this we can add script code that only compiles on later software versions only if the required version is used. This way, older software versions can still be used, the new feature will simply not be present there. So far, when using new script code functions, the code would not have compiled on newer versions effectively increasing the minimal required software version.


src/control/script_reader.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+  return script_code;
+}
+
+std::string ScriptReader::readFileContent(const std::string& filename)
+{
+  std::ifstream ifs;
+  ifs.open(filename);
+  std::string content;
+  std::ifstream file(filename);
```
urmahp: Remove

src/control/script_reader.cpp
```diff
@@ -0,0 +1,403 @@
...
+    if ((std::holds_alternative<double>(lhs) && std::holds_alternative<int>(rhs)))
+    {
+      return std::get<double>(lhs) == static_cast<double>(std::get<int>(rhs));
+    }
+    // Allow comparison between bool and int/double
+    if (std::holds_alternative<bool>(lhs))
+    {
+      if (std::holds_alternative<double>(rhs))
+      {
+        return std::abs((static_cast<double>(std::get<bool>(lhs)) - std::get<double>(rhs))) < 0.0001;
```
urmahp: Add a constant for the 0.0001 to increase readability

include/ur_client_library/control/script_reader.h
```diff
@@ -0,0 +1,128 @@
...
+  {
+    BlockType type;
+    bool condition_matched;  // Has any previous condition in this block matched?
+    bool should_render;      // Should this block render?
+    bool parent_render;      // Is the parent block rendering?
+  };
+
+  std::filesystem::path script_path_;
+
+  static std::string readFileContent(const std::string& filename);
```
urmahp:
```suggestion
  static std::string readFileContent(const std::string& filePath);
```
We are asking for the path to the file and not just the name

src/control/script_reader.cpp
```diff
@@ -0,0 +1,403 @@
...
+  std::string script_code = readFileContent(filename);
+
+  replaceVariables(script_code, data);
+  replaceConditionals(script_code, data);
+  replaceIncludes(script_code, data);
+
+  return script_code;
+}
+
+std::string ScriptReader::readFileContent(const std::string& filename)
```
urmahp: filePath instead of filename

src/control/script_reader.cpp
```diff
@@ -0,0 +1,403 @@
...
+void ScriptReader::replaceIncludes(std::string& script, const DataDict& data)
+{
+  std::regex include_pattern(R"(\{\%\s*include\s*['|"]([^'"]+)['|"]\s*\%\})");
+
+  std::smatch match;
+
+  // Replace all include patterns in the line
+  while (std::regex_search(script, match, include_pattern))
+  {
+    std::filesystem::path file_path(match[1]);
```
urmahp: It is a little misleading to call this file_path as it essentially is the file path from the directory where the main script is located, so in most cases, this will only consist of the filename, can we call it relative_file_path instead? 
```suggestion
    std::filesystem::path relative_file_path(match[1]);
```

doc/architecture/script_reader.rst
```diff
@@ -0,0 +1,132 @@
...
+
+- ``str``: A string value, e.g. "Hello World"
+- ``int``: An integer value, e.g. 42
+- ``double``: A floating point value, e.g. 3.14
+- ``bool``: A boolean value, e.g. ``true`` or ``false``
+- ``VersionInformation``: A version information value, e.g. ``VersionInformation::fromString("10.8.0")``
+
+Script code side
+~~~~~~~~~~~~~~~~
+
```
urmahp, line 82: Could we write just briefly how we add variables to the script, especially string variables, as they need to be encapsulated in quotes on the script side, that is the only thing I am missing from the documentation.

src/control/script_reader.cpp
```diff
@@ -0,0 +1,403 @@
...
+    }
+    else if (std::regex_search(line, match, elif_pattern))
+    {
+      if (!block_stack.empty())
+      {
+        BlockState& top = block_stack.top();
+        if (top.type == ELSE)
+          continue;
+        std::string condition = match[1];
+        bool result = data.count(condition) && std::holds_alternative<bool>(data.at(condition)) &&
```
urmahp: Maybe I have misunderstood something, but should we evaluate the expression here, since it is an elif statement the same way as it is does in the if statement.

src/control/script_reader.cpp
```diff
@@ -0,0 +1,403 @@
...
+    else if (std::regex_search(line, match, endif_pattern))
+    {
+      if (!block_stack.empty())
+      {
+        block_stack.pop();
+      }
+    }
+    else
+    {
+      if (block_stack.empty() ? true : block_stack.top().should_render)
```
urmahp: The above could be simplified to
```suggestion
      if (block_stack.empty() || block_stack.top().should_render)
```

src/control/script_reader.cpp
```diff
@@ -0,0 +1,403 @@
...
+      return data.at(variable_name) > value;
+    }
+    else if (comp_operator == "<=")
+    {
+      return data.at(variable_name) <= value;
+    }
+    else if (comp_operator == ">=")
+    {
+      return data.at(variable_name) >= value;
+    }
```
urmahp, line 384: Should we be explicit here and throw an exception if the comp operator doesn't match any of the provided values.

src/control/script_reader.cpp
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  script_code = output.str();
+}
+
+bool ScriptReader::evaluateExpression(const std::string& expression, const DataDict& data)
+{
+  const std::string trimmed = std::regex_replace(expression, std::regex("^\\s+|\\s+$"), "");
+  const std::vector<std::string> valid_operators = { "==", "!=", "<", ">", "<=", ">=" };
```
urmahp: This variable is not used
```suggestion
```

include/ur_client_library/ur/ur_driver.h
```diff
@@ -891,7 +892,7 @@ class UrDriver
     trajectory_interface_->registerDisconnectionCallback(fun);
   }
 
-  static std::string readScriptFile(const std::string& filename);
+  std::string readScriptFile(const std::string& filename);
```
urmahp: Not that I mind, but this is technically a breaking change. 
And also not really related to this PR, but is there a good reason that this function is public, I cannot see any use expect for reading the script file that is parsed through the config object.

src/ur/ur_driver.cpp
```diff
@@ -41,16 +43,16 @@
...
-static const std::string FORCE_MODE_SET_GAIN_SCALING_REPLACE("{{FORCE_MODE_SET_GAIN_SCALING_REPLACE}}");
+static const std::string BEGIN_REPLACE("BEGIN_REPLACE");
+static const std::string JOINT_STATE_REPLACE("JOINT_STATE_REPLACE");
+static const std::string TIME_REPLACE("TIME_REPLACE");
+static const std::string SERVO_J_REPLACE("SERVO_J_REPLACE");
+static const std::string SERVER_IP_REPLACE("SERVER_IP_REPLACE");
+static const std::string SERVER_PORT_REPLACE("SERVER_PORT_REPLACE");
+static const std::string TRAJECTORY_PORT_REPLACE("TRAJECTORY_SERVER_PORT_REPLACE");
+static const std::string SCRIPT_COMMAND_PORT_REPLACE("SCRIPT_COMMAND_SERVER_PORT_REPLACE");
+static const std::string FORCE_MODE_SET_DAMPING_REPLACE("FORCE_MODE_SET_DAMPING_REPLACE");
```
urmahp, line 54: This doesn't seem to be used can we remove it?

src/ur/ur_driver.cpp
```diff
@@ -41,16 +43,16 @@
...
+static const std::string FORCE_MODE_SET_GAIN_SCALING_REPLACE("FORCE_MODE_SET_GAIN_SCALING_REPLACE");
```
urmahp, line 55: Same as above

src/ur/ur_driver.cpp
```diff
@@ -41,16 +43,16 @@
...
-static const std::string JOINT_STATE_REPLACE("{{JOINT_STATE_REPLACE}}");
-static const std::string TIME_REPLACE("{{TIME_REPLACE}}");
-static const std::string SERVO_J_REPLACE("{{SERVO_J_REPLACE}}");
-static const std::string SERVER_IP_REPLACE("{{SERVER_IP_REPLACE}}");
-static const std::string SERVER_PORT_REPLACE("{{SERVER_PORT_REPLACE}}");
-static const std::string TRAJECTORY_PORT_REPLACE("{{TRAJECTORY_SERVER_PORT_REPLACE}}");
-static const std::string SCRIPT_COMMAND_PORT_REPLACE("{{SCRIPT_COMMAND_SERVER_PORT_REPLACE}}");
-static const std::string FORCE_MODE_SET_DAMPING_REPLACE("{{FORCE_MODE_SET_DAMPING_REPLACE}}");
-static const std::string FORCE_MODE_SET_GAIN_SCALING_REPLACE("{{FORCE_MODE_SET_GAIN_SCALING_REPLACE}}");
+static const std::string BEGIN_REPLACE("BEGIN_REPLACE");
```
urmahp, line 46: Would be nice with a name that resembles what it does, since this variable is replaced with the tool communication, but that would probably also be breaking the api if someone relies on a custom script.

src/ur/ur_driver.cpp
```diff
@@ -552,17 +527,13 @@ bool UrDriver::stopControl()
 
 std::string UrDriver::readScriptFile(const std::string& filename)
 {
-  std::ifstream ifs;
-  ifs.open(filename);
-  if (!ifs)
+  if (script_reader_ == nullptr)
```
urmahp: Can this happen? Isn't the script_reader initialized through the constructor

tests/test_script_reader.cpp
```diff
@@ -0,0 +1,408 @@
...
+  EXPECT_EQ(processed_script, "movej([1,2,3,4,5,6])");
+
+  std::remove(existing_script_file);
+  std::remove(existing_include_file);
+}
+
+TEST_F(ScriptReaderTest, ReplaceVariables)
+{
+  ScriptReader reader;
+  ScriptReader::DataDict data;
```
urmahp: Please ignore if you don't agree! But I prefer to have the variable definition just before they are used, so just above the readScriptFile.

tests/test_script_reader.cpp
```diff
@@ -0,0 +1,408 @@
...
+  // By default std::to_string will convert double to 6 decimal places
+  EXPECT_EQ(script, "movej([value1, 42, 6.280000, 0, 0, 0])\nlocal is_true = True\nlocal is_false = False\nThis is "
+                    "just a line without any replacement");
+  std::remove(existing_script_file);
+}
+
+TEST_F(ScriptReaderTest, VariableNotInDictThrowsError)
+{
+  ScriptReader reader;
+  ScriptReader::DataDict data;
```
urmahp: Same as above

tests/test_script_reader.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+  EXPECT_THROW(reader.readScriptFile(existing_script_file, data), urcl::UnknownVariable);
+  std::remove(existing_script_file);
+}
+
+TEST_F(ScriptReaderTest, ReplaceConditionals)
+{
+  ScriptReader reader;
+  ScriptReader::DataDict data;
+  data["is_logged_in"] = true;
```
urmahp: Same as above

tests/test_script_reader.cpp
```diff
@@ -0,0 +1,408 @@
...
+  // String not empty
+  EXPECT_TRUE(reader.evaluateExpression("S != ''", data));
+  EXPECT_TRUE(reader.evaluateExpression("S != \"\"", data));
+  EXPECT_FALSE(reader.evaluateExpression("S == \"\"", data));
+
+  // Provoke errors
+  // Non-existing operator
+  EXPECT_THROW(reader.evaluateExpression("X ~= 5", data), std::runtime_error);
+  EXPECT_THROW(reader.evaluateExpression("This is not an expression at all", data), std::runtime_error);
+  EXPECT_TRUE(reader.evaluateExpression("S != \"This is not an expression at all\"", data));
```
urmahp, line 351: It is intended to have this among all those throws or was it supposed to look different. Maybe like this?
```suggestion
  EXPECT_THROW(reader.evaluateExpression("S =! \"This is not an expression at all\"", data));
```

src/ur/ur_driver.cpp
urfeex: I wouldn't expect that, no. But if anyone changes the constructor ever that might happen and then we have a proper exception explaining the error.

include/ur_client_library/ur/ur_driver.h
urfeex: yeah, I had the same thought. The reason why this is public is for testability purposes as far as I can see. Now that I think about it, we could also leave the function as is. It isn't used inside the constructor, anyway and the functionality also doesn't change since using the script reader with an empty data dict is essentially the same as just reading the file. Should I add a deprecation warning instead that this will be removed in the next big change in 2027, instead?

src/ur/ur_driver.cpp
urfeex: I didn't necessarily want to change it inside this PR. It hasn't been used before, either, right?

src/ur/ur_driver.cpp
urfeex: same as above :-)

src/ur/ur_driver.cpp
urfeex: yes, I would consider this tremendously breaking. And it kind of does explain what it does, it's just not very explicit. It's the replacement at the beginning ;-) Back then we added that, we only had one or two substitutions as far as I can remember.

tests/test_script_reader.cpp
urfeex: No, that was intentional as far as I remember. But the check seems redundant, as we check strings with boolean operators at some other place, as well.

doc/architecture/script_reader.rst
urfeex: Would you agree with 76f8d01?

src/control/script_reader.cpp
urfeex: I think, you are absolutely right, I must have missed that when adding the expression evaluation. I will add another test with nested if/else blocks including more expressions.

urfeex: Fixed in c266999

include/ur_client_library/ur/ur_driver.h
urfeex: Resolved in 1d04975

src/control/script_reader.cpp
urfeex: The possible operators are part of the expression pattern, which is why I didn't add an exception here. It would not be possible to reach that code path.

doc/architecture/script_reader.rst
urmahp: Yes that is a good extension to the documentation.

include/ur_client_library/ur/ur_driver.h
```diff
@@ -891,6 +892,18 @@ class UrDriver
...
+   *
+   * This doesn't perform any substitutions on the file contents, but simply reads the file into a string.
+   *
+   * \deprecated This function isn't used. Please use the ScriptReader class instead.
+   *
+   * \param filename The name of the script file to read.
+   *
+   * \returns The content of the script file as a string.
+   */
+  [[deprecated("This function isn't used. Please use the ScriptReader class instead.")]]
```
urmahp: Should we write, when will remove the function?

src/ur/ur_driver.cpp
urmahp: No I don't recall that it has been used, but agree that it should be part of this PR, just stumbled across it while reviewing.

urmahp (approved):
Just one small comment regarding adding remove date to readScriptFile function other than that I am fine with merging this.

PR #348, opened 2025-06-25
Add torque_command statement only on supported software versions

urfeex:
This way all non-torque-control-related things should work on software versions not supporting torque control

ToDo:

- ~~Add tests for torque_control on different robot versions~~ We test that the not-including part works. For the include part we'll have to wait for a released URSim with torque_control support.

### DIFF
--- a/examples/script_command_interface.cpp
+++ b/examples/script_command_interface.cpp
@@ -78,8 +78,13 @@ void sendScriptCommands()
       g_my_robot->getUrDriver()->setPDControllerGains({ 500.0, 500.0, 300.0, 124.0, 124.0, 124.0 },
                                                       { 44.72, 44.72, 34.64, 22.27, 22.27, 22.27 });
     });
-    run_cmd("Setting max joint torques",
-            []() { g_my_robot->getUrDriver()->setMaxJointTorques({ 27.0, 27.0, 14.0, 4.5, 4.5, 4.5 }); });
+    auto robot_version = g_my_robot->getUrDriver()->getVersion();
+    if (robot_version > urcl::VersionInformation::fromString("10.10.0") ||
+        (robot_version.major == 5 && robot_version.minor >= 23))
+    {
+      run_cmd("Setting max joint torques",
+              []() { g_my_robot->getUrDriver()->setMaxJointTorques({ 27.0, 27.0, 14.0, 4.5, 4.5, 4.5 }); });
+    }
   }
   URCL_LOG_INFO("Script command thread finished.");
 }
--- a/examples/script_command_interface.cpp
+++ b/examples/script_command_interface.cpp
@@ -78,13 +78,9 @@ void sendScriptCommands()
       g_my_robot->getUrDriver()->setPDControllerGains({ 500.0, 500.0, 300.0, 124.0, 124.0, 124.0 },
                                                       { 44.72, 44.72, 34.64, 22.27, 22.27, 22.27 });
     });
-    auto robot_version = g_my_robot->getUrDriver()->getVersion();
-    if (robot_version > urcl::VersionInformation::fromString("10.10.0") ||
-        (robot_version.major == 5 && robot_version.minor >= 23))
-    {
-      run_cmd("Setting max joint torques",
-              []() { g_my_robot->getUrDriver()->setMaxJointTorques({ 27.0, 27.0, 14.0, 4.5, 4.5, 4.5 }); });
-    }
+    // The following will have no effect on PolyScope < 5.23 / 10.10
+    run_cmd("Setting max joint torques",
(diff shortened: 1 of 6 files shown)


examples/script_command_interface.cpp
```diff
@@ -78,8 +78,13 @@ void sendScriptCommands()
       g_my_robot->getUrDriver()->setPDControllerGains({ 500.0, 500.0, 300.0, 124.0, 124.0, 124.0 },
                                                       { 44.72, 44.72, 34.64, 22.27, 22.27, 22.27 });
     });
-    run_cmd("Setting max joint torques",
-            []() { g_my_robot->getUrDriver()->setMaxJointTorques({ 27.0, 27.0, 14.0, 4.5, 4.5, 4.5 }); });
+    auto robot_version = g_my_robot->getUrDriver()->getVersion();
```
urmahp: Is there a reason for this check, it is technically possible to call the script command in all versions, it just doesn't have any affect.

resources/external_control.urscript
```diff
@@ -760,8 +782,10 @@ thread script_commands():
           friction_compensation_enabled = True
         end
       elif command == SET_PD_CONTROLLER_GAINS:
+        {% if ROBOT_SOFTWARE_VERSION > v5.10.0 %}
```
urmahp, line 785: I am wondering whether we should add some sort of warning if this is called on a software version where it is not used as calling it wont have any affect, the same goes for friction compensation and max joint torques? This warning could be in the driver or the script command object.

src/ur/ur_driver.cpp
```diff
@@ -144,13 +144,23 @@ void UrDriver::init(const UrDriverConfiguration& config)
 
   const control::PDControllerGains pd_gains = control::getPdGainsFromRobotType(robot_type);
   std::stringstream pd_gains_ss;
-  pd_gains_ss << "struct(kp=" << pd_gains.kp << ", kd=" << pd_gains.kd << ")";
+  if (robot_version_ < urcl::VersionInformation::fromString("5.10.0"))
```
urmahp, line 150: Are we still support software version below 5.10? Would be better to check whether it is an cb3 robot?

urmahp (changes requested):

examples/script_command_interface.cpp
urfeex: no, that's a valid point.

resources/external_control.urscript
urfeex: yes, having the warning on the external side is definitively the better choice.

src/ur/ur_driver.cpp
urfeex: I have basically chosen 5.10 as structs are supported from there. Since torque control will never be available for CB3 I didn't bother checking that.

urmahp: It seems reasonable I also just realized that we support from 5.9.3 an onwards so it is a valid check.

include/ur_client_library/control/script_command_interface.h
```diff
@@ -226,6 +232,19 @@ class ScriptCommandInterface : public ReverseInterface
     SET_MAX_JOINT_TORQUES = 9,      ///< Set max joint torques
   };
 
+  void warnIfNotSupported(const VersionInformation& min_polyscope5, const VersionInformation& min_polyscopeX,
```
urmahp: Is there a reason for implementing this function in header? In my opinion I would put it into the cpp file.

include/ur_client_library/control/reverse_interface.h
```diff
@@ -62,6 +63,17 @@ enum class FreedriveControlMessage : int32_t
   FREEDRIVE_START = 1,  ///< Represents command to start freedrive mode.
 };
 
+struct ReverseInterfaceConfig
```
urmahp: This seems like a new preferred way to initialize the reverse interface, so wouldn't it make sense to add it to the master branch first instead of a feature branch. Or do we plan to merge torque control into the master branch after these changes as we can now use the version to determine whether the script functions are supported?

include/ur_client_library/control/script_command_interface.h
```diff
@@ -63,6 +64,11 @@ class ScriptCommandInterface : public ReverseInterface
    */
   ScriptCommandInterface(uint32_t port);
 
+  ScriptCommandInterface(const ReverseInterfaceConfig& config) : ReverseInterface(config)
```
urmahp: I would prefer the implementation of the constructor in the cpp file.

urmahp (changes requested):

include/ur_client_library/control/reverse_interface.h
urfeex: Hm, that's right. I will factor this out into a single PR to master.

urfeex:
I've rebased this on #351 merged into torque_control. @urmahp could you please have another look?

urmahp (approved):

Issue #352, opened 2025-07-05
[FEATURE NAME] Inquiry Regarding RTDE Control Interface API for Direct Torque Control


kwongeunwoo:
### Feature summary

Hello,

I am a student currently attempting direct torque control on a UR5e robot. I am working in the PolyScope 5.22.0 beta environment.
I am using the RTDE Control Interface API to send torque commands to the robot.

I have written a simple example code that sets all joint torques to zero using #include <ur_rtde/rtde_control_interface.h>.
However, when I run the code, the robot vibrates excessively even with the slightest touch.
I am reaching out to ask if there are any solutions or advice regarding this issue.

Below is the code I have written: 

Thank you for your attention.
==================================================================================
#include <ur_rtde/rtde_control_interface.h>
#include <ur_rtde/rtde_receive_interface.h>
#include <vector>
#include <thread>
#include <atomic>
#include <csignal>
#include <chrono>
#include <iostream>

using namespace ur_rtde;

std::atomic<bool> running(true);

void signal_handler(int signal) {
    if (signal == SIGINT) {
        std::cout << "\nCtrl+C pressed, exiting..." << std::endl;
        running = false;
    }
}

int main(int argc, char* argv[]) {
    std::cout << "Starting Torque Zero Command (C++)" << std::endl;

    RTDEControlInterface rtde_control("192.168.235.5", 500.0, RTDEControlInterface::FLAG_USE_EXT_UR_CAP, 50002);
    RTDEReceiveInterface rtde_receive("192.168.235.5", 500.0);

    if (!rtde_receive.isConnected()) {
        std::cerr << "[ERROR] RTDEReceiveInterface not connected!" << std::endl;
        return 1;
    }
    if (!rtde_control.isConnected()) {
        std::cerr << "[ERROR] RTDEControlInterface not connected!" << std::endl;
        return 1;
    }
    std::cout << "Successfully connected to robot!\n";

    std::signal(SIGINT, signal_handler);

    const double dt = 1.0 / 500.0; 

    std::cout << "Start torque zero command loop (Ctrl+C to exit)\n";
    while (running) {
        auto t_start = rtde_control.initPeriod();

        std::vector<double> tau_zero(6, 0.0); 
        rtde_control.torqueCommand(tau_zero, true);

        rtde_control.waitPeriod(t_start); 
    }

    rtde_control.stopScript();
    std::cout << "Torque zero finished.\n";
    return 0;
}
==================================================================================

[...]

PR #353, opened 2025-07-07
Add torque control example

urfeex:
Add an example using direct torque control.

This came up in #352

### DIFF
--- a/examples/torque_control.cpp
+++ b/examples/torque_control.cpp
@@ -59,6 +59,19 @@ int main(int argc, char* argv[])
     URCL_LOG_ERROR("Something in the robot initialization went wrong. Exiting. Please check the output above.");
     return 1;
   }
+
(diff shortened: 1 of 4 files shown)


examples/torque_control.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  bool passed_negative_part = false;
+  bool passed_positive_part = false;
+  URCL_LOG_INFO("Start moving the robot");
+  urcl::vector6d_t target_torques = { 2.0, 0, 0, 0, 0, 0 };
+
+  // Once RTDE communication is started, we have to make sure to read from the interface buffer, as
+  // otherwise we will get pipeline overflows. Therefor, do this directly before starting your main
+  // loop.
+  g_my_robot->getUrDriver()->startRTDECommunication();
```
SarveshMalladi, line 89: Im assuming that startRTDECommunication sets the bool and its unit tested

PR #354, opened 2025-07-07
Support optimove motions in InstructionExecutor

urfeex:
Add support for `optimovej` and `optimovel` in the `InstructionExecutor`.

### DIFF
--- a/tests/test_instruction_executor.cpp
+++ b/tests/test_instruction_executor.cpp
@@ -415,8 +415,8 @@ TEST_F(InstructionExecutorTest, optimovel_succeeds)
   // move to a feasible starting pose
   ASSERT_TRUE(executor_->moveJ({ -1.57, -1.6, 1.6, -0.7, 0.7, 0.2 }));
 
-  ASSERT_TRUE(executor_->optimoveL({ -0.203, 0.263, 0.559, 0.68, -1.083, -2.076 }, 0.4, 0.7, 0.1));
-  ASSERT_TRUE(executor_->optimoveL({ -0.203, 0.263, 0.459, 0.68, -1.083, -2.076 }, 1.0, 1.0, 0.1));
+  ASSERT_TRUE(executor_->optimoveL({ -0.203, 0.463, 0.559, 0.68, -1.083, -2.076 }, 0.4, 0.7, 0.1));
+  ASSERT_TRUE(executor_->optimoveL({ -0.203, 0.463, 0.459, 0.68, -1.083, -2.076 }, 1.0, 1.0, 0.1));
 }
 
 TEST_F(InstructionExecutorTest, optimovej_with_illegal_parameters_fails)


doc/examples/instruction_executor.rst
```diff
@@ -44,10 +44,11 @@ To run a sequence of motions, create an
...
 
 Each element in the motion sequence can be a different motion type. In the example, there are two
-``MoveJ`` motions, a ``MoveL`` motion and a ``MoveP`` motion. The primitives' parameters are directly forwarded to
-the underlying script functions, so the parameter descriptions for them apply, as well.
-Particularly, you may want to choose between either a time-based execution speed or an acceleration
-/ velocity parametrization for some move functions. The latter will be ignored if a time > 0 is given.
+``MoveJ`` motions, a ``MoveL`` motion, a ``MoveP`` motion, a ``OptimiveJ`` motion and a
+``OptimoveL`` motion. The primitives' parameters are directly forwarded to the underlying script
+functions, so the parameter descriptions for them apply, as well. Particularly, you may want to
+choose between either a time-based execution speed or an acceleration / velocity parametrization
```
urmahp:
```suggestion
choose between either a time-based execution or an acceleration / velocity parametrization
```

include/ur_client_library/control/motion_primitives.h
```diff
@@ -162,6 +176,40 @@ struct SplinePrimitive : public MotionPrimitive
...
+    type = MotionType::OPTIMOVEJ;
+    target_joint_configuration = target;
+    this->blend_radius = blend_radius;
+    this->acceleration = acceleration_fraction;
+    this->velocity = velocity_fraction;
+  }
+
+  urcl::vector6d_t target_joint_configuration;
+
+  bool validate() const override;
```
urmahp, line 188: For consistency move this function above the type definition.

src/control/motion_primitives.cpp
```diff
@@ -0,0 +1,158 @@
...
+  }
+  if (velocity < 0)
+  {
+    URCL_LOG_ERROR("Negative velocity passed to MoveJ primitive. This is not allowed.");
+    return false;
+  }
+  return true;
+}
+
+bool MoveLPrimitive::validate() const
```
urmahp: The validation seems to be the same for all motions types expect optimove, would it make sense to move this check to the parent function, since it doesn't break anything for the optimove check.

urmahp (changes requested):

src/control/motion_primitives.cpp
urfeex: I had that before #eb0ae05. However, splines also are different. We could move things back to the parent motion primitive, though and not call the base validation function from the spline primitive.

urmahp: Ahh I missed the splines, then I would probably leave it as it is.

urmahp (approved):

PR #356, opened 2025-07-09
REVERT_ME: Switch version checks to 5.22.0 as that't the beta version

urfeex:
Temporarily relax the version check for torque control features.

The torque control features will be available from 5.23.0 on. However, as we are currently running a public beta program, where the reported software version is 5.22.0 we should relax the version check until 5.23.0 is actually released.


urmahp (approved):

urfeex:
clang-tidy review says "All clean, LGTM! :+1:"

PR #358, opened 2025-07-09
Added registerTrajDoneCallback() for https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1341

mathias31415:
See https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1341#issuecomment-2931123202


urfeex (commented):
I think, a better approach might be to allow multiple callbacks registered in the driver. I'll make a PR for that.

PR #359, opened 2025-07-10
Add possibility to register multiple callbacks to ReverseInterface and TrajectoryPointInterface

urfeex:
This is an alternative to #358.

It still needs tests, hence the draft status.


urfeex:
Tests have been added by now, so I will mark this as ready. @mathias31415 as far as I can see, that implementation suits your usecase, right?

SarveshMalladi (approved):

PR #378, opened 2025-09-18
Only update force mode scaling and damping when needed

urrsk:
To avoid any impact of calling the scaling or damping methods if no change to these values are needed.


urfeex (approved):
That seems fair and tested that it works properly.

PR #379, opened 2025-09-19
Add const qualifier to get functions and changed map value retrieval to at() function

PabloDArandaR:
Currently getData functions are not const, although they shouldn't modify the data since they are retrieval functions.

It is a minor change I though about while using the library.


urfeex (approved):
Thank you, that's a valid approach.

PR #381, opened 2025-09-22
Direct torque control

urfeex:
This branch adds direct torque control to the client library. This is a sub-feature of the torque_control branch, basically not including the impedance-like PD controller.

This PR adds the functionality to command the robot by sending joint torques directly.


doc/architecture/reverse_interface.rst
```diff
@@ -84,4 +86,19 @@ meaning:
...
+
+.. _direct_torque_control_mode:
+
+Direct torque control mode
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Direct torque control mode is available since PolyScope version 5.23.0 / 10.10.0. It allows to command
+joint torques directly to the robot.
+
+.. note:: Target tprques are given **after** gravity compensation. A vector of zeros will hold the current position
```
urmahp:
```suggestion
.. note:: Target torques are given **after** gravity compensation. A vector of zeros will hold the current position
```

src/control/script_command_interface.cpp
```diff
@@ -226,6 +226,32 @@ bool ScriptCommandInterface::endToolContact()
   return server_.write(client_fd_, buffer, sizeof(buffer), written);
 }
 
+bool ScriptCommandInterface::setFrictionCompensation(const bool friction_compensation_enabled)
+{
+  // ToDo: Increase to 5.23.0 once released
+  robotVersionSupportsCommandOrWarn(urcl::VersionInformation::fromString("5.22.0"),
```
urmahp: 5.23 has been released 
```suggestion
  robotVersionSupportsCommandOrWarn(urcl::VersionInformation::fromString("5.23.0"),
```

src/control/script_command_interface.cpp
```diff
@@ -226,6 +226,32 @@ bool ScriptCommandInterface::endToolContact()
   return server_.write(client_fd_, buffer, sizeof(buffer), written);
 }
 
+bool ScriptCommandInterface::setFrictionCompensation(const bool friction_compensation_enabled)
+{
+  // ToDo: Increase to 5.23.0 once released
```
urmahp: Remove 
```suggestion
```

src/control/script_command_interface.cpp
urmahp: Add an if statement to check if it is the correct version or otherwise return false and not send the command to the robot?

urfeex:
I think, we should make sure that all tests  succeed before merging. It seems like we need to skip a couple of tests on "older" software versions.

PR #383, opened 2025-10-02
Set force mode parameters from config

domire8:
Due to the new way of calling force mode and keeping backward compatiblity, a few minor things got lost.

FYI @SprGrf


urfeex (approved):
Thank you for catching and fixing this!

PR #384, opened 2025-10-10
Fixed auto reconnection to the RTDE server.

urmahp:
When reconnecting to RTDE it requires you to set the input and output recipes again, this is now done properly.

### DIFF
--- a/include/ur_client_library/rtde/rtde_client.h
+++ b/include/ur_client_library/rtde/rtde_client.h
@@ -255,7 +255,7 @@ private:
 
   bool setupCommunication(const size_t max_num_tries = 0,
                           const std::chrono::milliseconds reconnection_time = std::chrono::seconds(10));
-  std::pair<bool, uint16_t> setProtocolVersion();
+  uint16_t negotiateProtocolVersion();
   bool queryURControlVersion();
   void setTargetFrequency();
   bool setupOutputs(const uint16_t protocol_version);
--- a/src/rtde/rtde_client.cpp
+++ b/src/rtde/rtde_client.cpp
@@ -146,9 +146,9 @@ bool RTDEClient::setupCommunication(const size_t max_num_tries, const std::chron
   }
   pipeline_->run();
 
-  std::pair<bool, uint16_t> protocol_version = setProtocolVersion();
+  uint16_t protocol_version = negotiateProtocolVersion();
   // Protocol version must be above zero
-  if (protocol_version.first == false)
+  if (protocol_version == 0)
   {
     return false;
   }
@@ -161,7 +161,7 @@ bool RTDEClient::setupCommunication(const size_t max_num_tries, const std::chron
     setTargetFrequency();
   }
 
-  is_rtde_comm_setup = is_rtde_comm_setup && setupOutputs(protocol_version.second);
+  is_rtde_comm_setup = is_rtde_comm_setup && setupOutputs(protocol_version);
 
   is_rtde_comm_setup = is_rtde_comm_setup && isRobotBooted();
 
@@ -172,7 +172,7 @@ bool RTDEClient::setupCommunication(const size_t max_num_tries, const std::chron
   return is_rtde_comm_setup;
 }
 
-std::pair<bool, uint16_t> RTDEClient::setProtocolVersion()
+uint16_t RTDEClient::negotiateProtocolVersion()
 {
   uint16_t protocol_version = MAX_RTDE_PROTOCOL_VERSION;
   while (protocol_version > 0)
@@ -187,7 +187,7 @@ std::pair<bool, uint16_t> RTDEClient::setProtocolVersion()
     if (!stream_.write(buffer, size, written))
     {
       URCL_LOG_ERROR("Sending protocol version query to robot failed");
-      return { false, 0 };
+      return 0;
     }
 
     while (num_retries < MAX_REQUEST_RETRIES)
@@ -196,7 +196,7 @@ std::pair<bool, uint16_t> RTDEClient::setProtocolVersion()
       if (!pipeline_->getLatestProduct(package, std::chrono::milliseconds(1000)))
       {
         URCL_LOG_ERROR("failed to get package from RTDE interface");
-        return { false, 0 };
+        return 0;
       }
       if (rtde_interface::RequestProtocolVersion* tmp_version =
               dynamic_cast<rtde_interface::RequestProtocolVersion*>(package.get()))
@@ -205,7 +205,7 @@ std::pair<bool, uint16_t> RTDEClient::setProtocolVersion()
         {
           URCL_LOG_INFO("Negotiated RTDE protocol version to %hu.", protocol_version);
           parser_.setProtocolVersion(protocol_version);
-          return { true, protocol_version };
+          return protocol_version;
         }
         break;
       }
@@ -224,7 +224,7 @@ std::pair<bool, uint16_t> RTDEClient::setProtocolVersion()
   }
   URCL_LOG_ERROR("Protocol version for RTDE communication could not be established. Robot didn't accept any of "
                  "the suggested versions.");
-  return { false, 0 };
+  return 0;
 }
 
 bool RTDEClient::queryURControlVersion()


include/ur_client_library/comm/pipeline.h
```diff
@@ -366,13 +369,13 @@ class Pipeline
    */
   void stop()
   {
-    if (!running_)
+    if (threads_stopped_)
```
urfeex: Could you please elaborate why you added this? It seems almost redundant with `running_`, but if the producer/consumer threads fail, we set `running_` to `false` there, so you wanted to exclude that. I guess, in order to make sure we still stop and join both, the producer and consumer threads on a `stop()` call.

Couldn't we instead always make the necessary calls in that function? E.g. `producer_.stopProducer()` should always work. Thread joining should also not be the issue. So basically, the only thing we would like to to only once is calling the notifier and maybe the log output, right?

src/rtde/rtde_client.cpp
```diff
@@ -89,93 +91,87 @@ bool RTDEClient::init(const size_t max_connection_attempts, const std::chrono::m
     return true;
   }
 
+  prod_->setRTDEReconnectionCallback(std::bind(&RTDEClient::reconnectCallback, this));
+  std::cout << "what is happening here" << std::endl;
```
urfeex: What do I know? :-D

src/rtde/rtde_client.cpp
```diff
@@ -89,93 +91,87 @@ bool RTDEClient::init(const size_t max_connection_attempts, const std::chrono::m
...
+
   unsigned int attempts = 0;
   std::stringstream ss;
-  while (attempts < max_initialization_attempts)
+
+  // A running pipeline is needed inside setup.
+  pipeline_->init(max_connection_attempts, reconnection_timeout);
+  pipeline_->run();
```
urfeex: Why did you move this out of the `setupCommunication` method? It seems odd, especially since we're still calling it every time before calling `setupCommunication`.

src/rtde/rtde_client.cpp
```diff
@@ -89,93 +91,87 @@ bool RTDEClient::init(const size_t max_connection_attempts, const std::chrono::m
...
+  is_rtde_comm_setup = is_rtde_comm_setup && isRobotBooted();
+
+  return is_rtde_comm_setup && setupInputs();
+}
+
+uint16_t RTDEClient::setProtocolVersion()
+{
   uint16_t protocol_version = MAX_RTDE_PROTOCOL_VERSION;
-  while (!negotiateProtocolVersion(protocol_version) && client_state_ == ClientState::INITIALIZING)
+  while (!negotiateProtocolVersion(protocol_version))
```
urfeex: There seems to be some duplication with handling errors in the protocol negotiation process. However, I think that adding a `setProtocolVersion()` method that basically calls `negotiateProtocolVersion()` makes it any better. Could we rather merge the logic from both functions to `negotiateProtocolVersion()`?

src/rtde/rtde_client.cpp
```diff
@@ -269,7 +261,27 @@ void RTDEClient::queryURControlVersion()
...
+  {
+    // Default to maximum frequency
+    target_frequency_ = max_frequency_;
+  }
+  else if (target_frequency_ <= 0.0 || target_frequency_ > max_frequency_)
+  {
+    // Target frequency outside valid range
+    throw UrException("Invalid target frequency of RTDE connection");
+  }
+  return true;
```
urfeex: Since there is no path returning `false` this function could be void.

src/rtde/rtde_client.cpp
urmahp: haha thought I removed all my debugging statements, but apparently I forgot one :D

include/ur_client_library/comm/pipeline.h
urmahp: I also did just remove it initially as we cannot start the pipeline again unless the threads have been joined. The problem is actually in one of the tests where it destroys the producer object before destroying the pipeline and then we end up with an exception because the producer no longer exists and we try to call producer_.stopProducer(). I could also fix the test instead?

src/rtde/rtde_client.cpp
urmahp: All the methods in setting up the communication is becoming very large and hard to read, which is why i didn't merge the two methods, but if you prefer to merge then I could do that?

src/rtde/rtde_client.cpp
urmahp: Yes that is true, I did to be able to do.
`
is_rtde_comm_setup = is_rtde_comm_setup && setTargetFrequency();
`
instead off
`
if (is_rtde_comm_setup) {
  setTargetFrequency();
}
`
But I can go ahead and change it.

include/ur_client_library/comm/pipeline.h
urfeex: Yeah, that seems like the test isn't doing it properly. If the pipeline needs the producer, then we shouldn't destroy the producer before destroying the pipeline.

src/rtde/rtde_client.cpp
urfeex: I understand that and I am very much in favor of your refactoring. I just think that here we now have two functions that are named very similarly and `setProtocolVersion` does indeed also do (internally) negotiate the protocol version.

urmahp: I will try to merge them and then we can have a look.

src/rtde/rtde_client.cpp
urmahp: Because of some of the testing I did, where it was still running when disconnecting. But since it is no longer it should be moved back.

include/ur_client_library/comm/producer.h
```diff
@@ -124,9 +125,28 @@ class URProducer : public IProducer<T>
...
+        {
+          URCL_LOG_WARN("Failed to read from RTDE stream, invoking on reconnect callback and stopping the producer");
+          on_rtde_reconnect_cb_();
+        }
+        else
+        {
+          URCL_LOG_ERROR("Failed to read from RTDE stream without a reconnect handler stopping the producer");
+        }
+        return false;
+      }
```
urfeex: I didn't really like the special handling of an RTDE producer here. Couldn't we generalize this by checking whether there is a reconnection callback registered? If none then we simply do not reconnect and if yes, trigger reconnect independent of the stream type. Sorry that I just come up with this now, but that seems cleaner to me.

urmahp: Reconnecting to the stream, will just not have any affect if it is a RTDE stream, but I will change the check.

include/ur_client_library/comm/stream.h
```diff
@@ -30,6 +30,7 @@ namespace urcl
 {
 namespace comm
 {
+
```
urfeex:
```suggestion
```

include/ur_client_library/rtde/rtde_client.h
```diff
@@ -231,6 +231,10 @@ class RTDEClient
   comm::INotifier notifier_;
   std::unique_ptr<comm::Pipeline<RTDEPackage>> pipeline_;
   RTDEWriter writer_;
+  bool reconnecting_;
+  bool stop_reconnection_;
```
urfeex:
```suggestion
  std::atomic<bool> reconnecting_;
  std::atomic<bool> stop_reconnection_;
```

src/rtde/rtde_client.cpp
```diff
@@ -93,147 +103,131 @@ bool RTDEClient::init(const size_t max_connection_attempts, const std::chrono::m
...
   }
-
-  // We finished communication for now
-  pipeline_->stop();
-  client_state_ = ClientState::INITIALIZED;
+  return is_rtde_comm_setup;
 }
 
-bool RTDEClient::negotiateProtocolVersion(const uint16_t protocol_version)
+std::pair<bool, uint16_t> RTDEClient::setProtocolVersion()
```
urfeex: It seems like we are only interested in the boolean return value, so should we reduce the signature to that one?

urfeex (commented):
Just some nitpicky suggestions left that I can commit mys

urfeex (approved):

PR #388, opened 2025-10-10
Rtde external ft

urfeex:
Add support for sending external force-torque measurements through RTDE.

This can be especially useful if you want to inject fake measurements into a URSim simulated robot.

There's also an example with a simple TUI doing exactly that.

### DIFF
--- a/doc/examples/external_fts_through_rtde.rst
+++ b/doc/examples/external_fts_through_rtde.rst
@@ -39,7 +39,7 @@ The robot initialization is handed off to the ``ExampleRobotWrapper`` class:
 
 .. literalinclude:: ../../examples/external_fts_through_rtde.cpp
    :language: c++
-   :caption: external_fts_through_rtde/freedrive_example.cpp
+   :caption: external_fts_through_rtde/external_fts_through_rtde.cpp
    :linenos:
    :lineno-match:
    :start-at: g_my_robot = std::make_unique
@@ -47,15 +47,15 @@ The robot initialization is handed off to the ``ExampleRobotWrapper`` class:
 
 In order to use the RTDE input for the FTS, we have to enable it first by calling the
 ``ft_rtde_input_enable`` function. This is done by sending a script to the robot through the
-``sendScript()`` method of the ``UrDriver`` class.
+``ftRtdeInputEnable()`` method of the ``UrDriver`` class.
 
 .. literalinclude:: ../../examples/external_fts_through_rtde.cpp
    :language: c++
-   :caption: external_fts_through_rtde/freedrive_example.cpp
+   :caption: external_fts_through_rtde/external_fts_through_rtde.cpp
    :linenos:
    :lineno-match:
    :start-at: // Enable using the force-torque
-   :end-at: g_my_robot->getPrimaryClient()->sendScript("ft_rtde_input_enable(True)");
+   :end-at: g_my_robot->getUrDriver()->ftRtdeInputEnable(true);
 
 RTDE communication is moved to a separate thread in order to use the main thread for handling
 keyboard input commands.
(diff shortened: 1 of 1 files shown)


doc/examples/external_fts_through_rtde.rst
```diff
@@ -0,0 +1,165 @@
...
+   :lineno-match:
+   :start-at: if (g_FT_VEC_MUTEX.try_lock())
+   :end-at: }
+
+Note that the ``try_lock()`` method is used in the RTDE thread to avoid blocking the thread if the
+tui thread is currently holding the lock. The new values will be available in a later cycle then.
+
+.. note::
+   The values reported from the robot are rotated to the base coordinate system of the robot.
+   Hence, they will not be the same as the ones sent to the robot. Align the robot's TCP the same
```
urmahp: For them to be the same, align the TCP frame with robots base frame.

examples/external_fts_through_rtde.cpp
```diff
@@ -0,0 +1,252 @@
...
+std::unique_ptr<urcl::ExampleRobotWrapper> g_my_robot;
+std::atomic<bool> g_RUNNING = true;
+urcl::vector6d_t g_FT_VEC{ 0, 0, 0, 0, 0, 0 };
+std::mutex g_FT_VEC_MUTEX;
+
+using namespace urcl;
+
+void ftInputTui()
+{
+  const std::string instructions = "Press x,y,z to increase the respective axes, 0 for reset, q for quit.";
```
urmahp:
```suggestion
  const std::string instructions = "Press x,y,z to increase the respective translational axes, a,b,c to increase the rotational axes, 0 for reset and q for quit.";
```

include/ur_client_library/comm/package_serializer.h
```diff
@@ -98,6 +122,24 @@ class PackageSerializer
     return val.size();
   }
 
+  /*!
+   * \brief A serialization method for std::arrays (works for urcl::vector6d_t, etc).
+   *
+   * \param buffer The buffer to write the serialization into.
+   * \param val The string to serialize.
```
urmahp:
```suggestion
   * \param val The array to serialize.
```

include/ur_client_library/comm/package_serializer.h
```diff
@@ -71,6 +75,7 @@ class PackageSerializer
    */
   static size_t serialize(uint8_t* buffer, double val)
   {
+    static_assert(sizeof(double) == sizeof(uint64_t), "Unexpected double size");
```
urmahp, line 78: Is there a reason for this check?

include/ur_client_library/comm/package_serializer.h
```diff
@@ -79,6 +84,25 @@ class PackageSerializer
...
+   * \brief A serialization method for float values.
+   *
+   * \param buffer The buffer to write the serialization into.
+   * \param val The value to serialize.
+   *
+   * \returns Size in byte of the serialization.
+   */
+  static size_t serialize(uint8_t* buffer, float val)
+  {
+    static_assert(sizeof(float) == sizeof(uint32_t), "Unexpected float size");
```
urmahp, line 97: Is there a reason for this check?

include/ur_client_library/comm/package_serializer.h
urfeex: I wasn't sure whether this is guaranteed on all platforms. It doesn't hurt, since this is only a compile time check. I am not aware of any architecture / software platform where this isn't given, though.

include/ur_client_library/comm/package_serializer.h
urfeex: Same as above

include/ur_client_library/comm/package_serializer.h
urmahp: Lets keep the check then.

include/ur_client_library/comm/package_serializer.h
urmahp: Lets keep the check

urmahp (approved):

PR #389, opened 2025-10-10
Refactor Rtde client test

URJala:
The exhaustive rtde output recipe is now generated at build time, and the test will fail if the controller does not know all fields. The test will also fail if using the wrong ursim version


tests/resources/generate_rtde_outputs.py
urfeex, line 1: I think it would be best, if we queried all possible fields from https://docs.universal-robots.com/tutorials/communication-protocol-tutorials/rtde-guide.html

This way, that test would fail if we forgot to add a new field to the library.

Edit: On the other hand, that might be a different test, after all.

tests/resources/generate_rtde_outputs.py
```diff
@@ -0,0 +1,29 @@
+import pathlib
```
urfeex: Please  add a license header to that file.

PR #401, opened 2025-11-06
Add bit for 3PE stoppage to UrRtdeSafetyStatusBits

### LINKED ISSUE #394: UrRtdeSafetyStatusBits does not have an entry for 3PE signal
### Affected Client Library version(s)

2.4.0

### What combination of platform is the Client Library running on.

Linux

### How did you install the Client Library

Build the library from source

### Which robot platform is the library connected to.

Real robot

N/A

### How are you using the Client Library

Others

### Issue details

The documentation for the `safety_status_bits` RTDE field indicates 12 bits (0-11), with the last bit being "is 3PE input active". However, `UrRtdeSafetyStatusBits` only defines 11 symbols, and lacks an entry for 3PE: https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/173e3c386e4995651e118024526a0040cd510821/include/ur_client_library/rtde/rtde_client.h#L67-L77.

Note that the `SafetyStatus` _enum_ does have 3PE relevant values: https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/173e3c386e4995651e118024526a0040cd510821/include/ur_client_library/ur/datatypes.h#L67.

So, I think this is just an oversight, but it would be nice if the `Bits` type included this information.

I'm happy to put together a PR for this if there is agreement that this issue should be fixed.

### Relevant log output

```shell

```

### Accept Public visibility

- [x] I agree to make this context public

acmorrow:
Fixes #394.

I think this is really all that is required, unless you have unit tests that you need, but I didn't find any results in the codebase for other symbols from this enumeration.

### DIFF
--- a/include/ur_client_library/rtde/rtde_client.h
+++ b/include/ur_client_library/rtde/rtde_client.h
@@ -75,7 +75,7 @@ enum class UrRtdeSafetyStatusBits
   IS_VIOLATION = 8,
   IS_FAULT = 9,
   IS_STOPPED_DUE_TO_SAFETY = 10,
-  IS_SYSTEM_THREE_POSITION_ENABLING_STOPPED = 11
+  IS_3PE_INPUT_ACTIVE = 11
 };
 
 enum class ClientState


include/ur_client_library/rtde/rtde_client.h
```diff
@@ -74,7 +74,8 @@ enum class UrRtdeSafetyStatusBits
   IS_EMERGENCY_STOPPED = 7,
   IS_VIOLATION = 8,
   IS_FAULT = 9,
-  IS_STOPPED_DUE_TO_SAFETY = 10
+  IS_STOPPED_DUE_TO_SAFETY = 10,
+  IS_SYSTEM_THREE_POSITION_ENABLING_STOPPED = 11
```
urfeex: I would suggest following the RTDE documentation regarding the naming:
```suggestion
  IS_3PE_INPUT_ACTIVE = 11
```

acmorrow: Thanks @urfeex for the feedback.

So, I picked this name by analogy with what I found in `SafetyStatus`: `SafetyStatus::ROBOT_EMERGENCY_STOP` goes to `UrRtdeSafetyStatusBits::IS_ROBOT_EMERGENCY_STOPPED`, so I analogized `SafetyStatus::SYSTEM_THREE_POSITION_ENABLING_STOP` to the proposed `UrRtdeSafetyStatusBits::IS_SYSTEM_THREE_POSITION_ENABLING_STOPPED`. Just let me know which you prefer. If you still prefer the one you proposed here over the analogized one, I'll just accept your edit and you can merge.

urfeex: Yes, I figured that's how you picked the name. However, they don't refer to the same thing exactly:

- There is also TP_THREE_POSITION_ENABLING_STOP. I don't know what the dashboard reports as SYSTEM_THREE_POSITION_ENABLING_STOP, but the RTDE field we're talking about is a boolean AND of the teach pendant 3pe and an external 3pe depending on which is present.
- I also don't know whether the STOPPED one from the dashboard server is actually the inverse of the INPUT_ACTIVE

We always try to be as close to the public documentation as possible to avoid uncertainties.

acmorrow: @urfeex - OK, I applied your suggested edit.

PR #406, opened 2025-11-14
Add 10.11.0 to ci matrix

urfeex:


urfeex:
This is now implemented in #435

PR #408, opened 2025-11-14
Added support for std::vector strings in UrDriver

PabloDArandaR:
Since it is a functionality allowed by the RTDEClient, seems reasonable to also add it to the UrDriver.

### DIFF
--- a/.github/workflows/check_links.yml
+++ b/.github/workflows/check_links.yml
@@ -10,7 +10,7 @@ jobs:
   check_links:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Restore lychee cache
         id: restore-cache
         uses: actions/cache/restore@v4
@@ -34,6 +34,7 @@ jobs:
             --exclude 'releases/download/v\$%7BURCAP_VERSION%7D/externalcontrol-\$%7BURCAP_VERSION%7D\.jar'
             --exclude '^http://rosin-project\.eu'
             --exclude '2013181\/gdb-causes-sem-wait-to-fail-with-eintr-error'
+            --exclude 'https:\/\/lists\.apple\.com\/archives\/darwin-kernel\/2009\/Apr\/msg00010\.html'
             --max-concurrency 1
             './**/*.md' './**/*.html' './**/*.rst' './**/*.cpp' './**/*.h' './**/*.py'
       - name: Save lychee cache
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
             PROGRAM_FOLDER: 'tests/resources/dockerursim/programs/polyscopex'
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: start ursim
         run: |
           scripts/start_ursim.sh -m $ROBOT_MODEL -v $URSIM_VERSION -p $PROGRAM_FOLDER -d
@@ -109,7 +109,7 @@ jobs:
   test_start_ursim:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: '3.3'  # Not needed with a .ruby-version, .tool-versions or mise.toml
--- a/.github/workflows/industrial-ci.yml
+++ b/.github/workflows/industrial-ci.yml
@@ -35,7 +35,7 @@ jobs:
     env:
       DOCKER_RUN_OPTS: '-v /var/run/docker.sock:/var/run/docker.sock --network ursim_net'
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - run: docker network create --subnet=192.168.56.0/24 ursim_net
         if: ${{ !env.ACT }}
       - uses: 'ros-industrial/industrial_ci@master'
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,7 +15,7 @@ jobs:
     name: pre-commit
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - uses: actions/setup-python@v6
       with:
         python-version: 3.10.4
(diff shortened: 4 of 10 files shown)


PabloDArandaR:
@urfeex Okay, I added some tests to the UrDriver to use the new functionality of the input parsing. Since they are not affecting each other, I made only 1 test for each of the following cases since the input/output are handled independently: only recipe files, only recipe vectors, and both at the same time.

include/ur_client_library/example_robot_wrapper.h
```diff
@@ -74,7 +74,8 @@ class ExampleRobotWrapper
    * communicating to the driver's reverse interface and trajectory interface.
    */
   ExampleRobotWrapper(const std::string& robot_ip, const std::string& output_recipe_file,
-                      const std::string& input_recipe_file, const bool headless_mode = true,
+                      const std::string& input_recipe_file, const std::vector<std::string> output_recipe,
+                      const std::vector<std::string> input_recipe, const bool headless_mode = true,
```
urfeex, line 97: This is definitively a breaking API change, which you also experienced changing all of the examples. May I suggest adding a second constructor accepting the vectors instead of the filenames? Alternatively, add this signature as second constructor and maybe deprecate the existing one.

PabloDArandaR: We can do that, but one particular case won't be tested (both the files and the vectors are defined). Since this is only used for testing, I thought about adding a constructor that  supports both inputs simultaneously (besides the current constructor). But to keep it consistent, I will add a third constructor that only has the vectors of strings

urfeex: I know, I did suggest the deprecation myself, but I think, we don't want to go down that route. Having two constructors, one with recipe files and one with recipe vectors should be sufficient. Then users can pick. Please also leave the examples as is for the scope of this PR.

For testing both being defined we don't need the example wrapper, the UrDriver can be used directly.

src/ur/ur_driver.cpp
```diff
@@ -684,4 +695,44 @@ std::deque<urcl::primary_interface::ErrorCode> UrDriver::getErrorCodes()
...
+{
+  bool use_output_file = true;
+  if (config.output_recipe_file.empty() && config.output_recipe.size() == 0)
+    throw UrException("Neither output recipe file nor output recipe have been defined");
+  else if (!config.output_recipe_file.empty() && config.output_recipe.size() != 0)
+    URCL_LOG_WARN("Both output recipe file and output recipe vector are  used. Defaulting to output recipe file");
+  else if (config.output_recipe_file.empty())
+    use_output_file = false;
+  if (use_output_file && !std::filesystem::exists(config.output_recipe_file))
+    throw UrException("Output recipe file does not exist: " + config.output_recipe_file);
```
urfeex:
```suggestion
  auto output_recipe = config.output_recipe;
  if (config.output_recipe_file.empty() && config.output_recipe.size() == 0)
  {
    throw UrException("Neither output recipe file nor output recipe have been defined. An output recipe is required.");
  }
  if (!config.output_recipe_file.empty())
  {
     if (config.output_recipe.size() != 0)
    URCL_LOG_WARN("Both output recipe file and output recipe vector are  used. Defaulting to output recipe vector");
  }
  else
  {
    output_recipe = rtde_interface::RTDEClient::readRecipe(config.output_recipe_file);
  }
```

src/ur/ur_driver.cpp
```diff
@@ -684,4 +695,44 @@ std::deque<urcl::primary_interface::ErrorCode> UrDriver::getErrorCodes()
...
+
+  bool use_input_file = true;
+  if (config.input_recipe_file.empty() && config.input_recipe.size() == 0)
+    throw UrException("Neither input recipe file nor input recipe have been defined");
+  else if (!config.input_recipe_file.empty() && config.input_recipe.size() != 0)
+    URCL_LOG_WARN("Both input recipe file and input recipe vector are  used. Defaulting to input recipe file");
+  else if (config.input_recipe_file.empty())
+    use_input_file = false;
```
urfeex:
```suggestion
  auto input_recipe = config.input_recipe;
  if (!config.input_recipe_file.empty())
  {
    if (config.input_recipe.size() != 0)
    {
      URCL_LOG_WARN("Both input recipe file and input recipe vector are  used. Defaulting to input recipe vector.");
    }
    else
    {
      input_recipe = rtde_interface::RTDEClient::readRecipe(config.input_recipe_file);
    }
```

src/ur/ur_driver.cpp
```diff
@@ -684,4 +695,44 @@ std::deque<urcl::primary_interface::ErrorCode> UrDriver::getErrorCodes()
...
+  {
+    auto input_recipe = config.input_recipe;
+    auto output_recipe = config.output_recipe;
+    if (use_input_file)
+      input_recipe = rtde_interface::RTDEClient::readRecipe(config.input_recipe_file);
+    if (use_output_file)
+      output_recipe = rtde_interface::RTDEClient::readRecipe(config.output_recipe_file);
+
+    rtde_client_.reset(new rtde_interface::RTDEClient(robot_ip_, notifier_, output_recipe, input_recipe));
+  }
```
urfeex:
```suggestion
    rtde_client_.reset(new rtde_interface::RTDEClient(robot_ip_, notifier_, output_recipe, input_recipe));
  }
```

urfeex (commented):
I think we can simplify the logic by always using the vector-based constructor internally. Also note, that an empty / no input recipe is allowed.

You seem to have removed the tests. Adding tests to the `UrDriverTest` suite should be quite straightforward.

src/rtde/rtde_client.cpp
```diff
@@ -664,7 +664,7 @@ bool RTDEClient::sendPause()
   throw UrException(ss.str());
 }
 
-std::vector<std::string> RTDEClient::readRecipe(const std::string& recipe_file) const
+std::vector<std::string> RTDEClient::readRecipe(const std::string& recipe_file)
```
PabloDArandaR, line 667: Why make it  non-const? The function is currently meant to parse the file, not to change the internal state of the client, isn't it?

urfeex: Basically, because it's a static function now (which in fact you changed yourself, I believe :-)), which makes much more sense.

PabloDArandaR: Indeed, I did, long week 😅

PR #419, opened 2026-01-21
Refactor rtde parsing

urfeex:
This is a refactor of RTDE communication effectively improving two things:

- No more heap memory allocations during communication once everything is setup
- Reduced parsing and access times of RTDE data that reduces the time of read and send operations

There's currently some documentation and testing missing, hence the draft status.

---

> [!NOTE]
> **High Risk**
> Touches core RTDE networking/parsing/IO concurrency and changes public APIs with deprecations; regressions could impact real-time data integrity, reconnection behavior, and write semantics under load.
> 
> **Overview**
> **RTDE read path is refactored to avoid per-cycle heap allocations and reduce overhead.** `RTDEClient` no longer relies on the `Pipeline` for reads; it uses `URProducer` directly plus a double-buffered background read thread, and introduces `start(bool read_packages_in_background)` with new `getDataPackage(DataPackage&/unique_ptr&, timeout)` and `getDataPackageBlocking(...)` APIs while deprecating the allocating `getDataPackage(timeout)` overload.
> 
> **RTDE parsing and packages are updated to support reuse and stricter validation.** `Parser`/`IProducer` add single-product parse/tryGet overloads, `RTDEParser` is split into a new `rtde_parser.cpp` with a prealloc-friendly parse path (vector-based parse deprecated), `RTDEPackage` exposes `getType()`, and `DataPackage` switches from an `unordered_map` to an ordered vector store, initializes empty data on construction, validates recipe keys (throwing new `RTDEInvalidKeyException`), adds type checking on `setData`, and improves `toString()` formatting.
> 
> **RTDE write path is refactored for fewer allocations and batching.** `RTDEWriter` replaces the moodycamel queue with double-buffered storage + condition variable, adds `sendPackage()` for sending a fully-prepared package, preallocates register key strings, and resets mask fields after sending; it now throws on `init()`/`setInputRecipe()` while running.
> 
> **API consumers, docs, and tests are migrated to the new allocation-free patterns.** Examples and documentation now show preallocating `DataPackage` and using the new read modes; `ExampleRobotWrapper` removes its custom RTDE consumer thread and forwards the background-read choice to `UrDriver`. Tests are expanded/updated for background vs blocking read, reconnection, invalid keys/conflicts, `DataPackage` parsing/toString, and parser single-product mode.
> 
> **CI and tooling tweaks:** workflow disables URSim port forwarding, gates coverage uploads on successful build, shortens example sleep between runs, and improves fake RTDE server thread shutdown safety.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4e233954621dc1d3bfa7b8fad236209eb054bee3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/rtde/rtde_client.cpp
+++ b/src/rtde/rtde_client.cpp
@@ -848,7 +848,6 @@ RTDEWriter& RTDEClient::getWriter()
 void RTDEClient::reconnect()
 {
   URCL_LOG_INFO("Reconnecting to the RTDE interface");
-  reconnecting_ = true;
   // Locking mutex to ensure that calling getDataPackage doesn't influence the communication needed for reconfiguring
   // the RTDE connection
   std::lock_guard<std::mutex> lock(reconnect_mutex_);
@@ -943,6 +942,7 @@ void RTDEClient::reconnectCallback()
   {
     reconnecting_thread_.join();
   }
+  reconnecting_ = true;
   reconnecting_thread_ = std::thread(&RTDEClient::reconnect, this);
 }
 


examples/rtde_client.cpp
```diff
@@ -84,7 +83,10 @@ int main(int argc, char* argv[])
   // Once RTDE communication is started, we have to make sure to read from the interface buffer, as
   // otherwise we will get pipeline overflows. Therefor, do this directly before starting your main
   // loop.
-  my_client.start();
+  my_client.start(false);
+  const std::string key = "target_speed_fraction";
```
urrsk: variable should have a more explaning name and maybe also in UPPERCAST letters as it is a constant

urfeex: done in the latest version.

doc/architecture/rtde_client.rst
```diff
@@ -28,27 +45,32 @@ outputs. Please refer to the `RTDE
...
-   following will work
+   rtde_interface::RTDEClient my_client(
+     ROBOT_IP,
+     notifier,
+     {"timestamp", "actual_q"},
+     {"speed_slider_mask", "speed_slider_fraction"}
+   );
+
+.. note::
+   Remember, that ``timestamp`` always has to be part of the output recipe.
```
urrsk:
```suggestion
   ``timestamp`` will always be a part of the output recipe and will be added afterwards, if not defined. As the ``timestamp`` used for verifying the connectivity.
```

include/ur_client_library/exceptions.h
```diff
@@ -272,5 +272,51 @@ class UnexpectedResponse : public UrException
...
+  explicit RTDEInputConflictException() : std::runtime_error("RTDE Output Conflict Exception")
+  {
+  }
+
+  explicit RTDEInputConflictException(const std::string& key)
+    : std::runtime_error("RTDE Output Conflict for key: " + key), key_(key)
+  {
+    message_ = "Variable '" + key_ +
+               "' is currently controlled by another RTDE client. The input recipe can't be used as "
+               "configured";
```
urrsk: Maybe we here also should give a hint that the IO's also are used by the fieldbuses?!

src/rtde/rtde_client.cpp
```diff
@@ -857,5 +922,61 @@ void RTDEClient::reconnectCallback()
...
+      }
+      else if (data_buffer1_->getType() == PackageType::RTDE_TEXT_MESSAGE)
+      {
+        URCL_LOG_INFO(data_buffer1_->toString().c_str());
+      }
+    }
+    else
+    {
+      auto period = std::chrono::duration<double>(1.0 / target_frequency_);
+      std::this_thread::sleep_for(period);
```
urrsk, line 1010: Maybe we should sleep less time than the target frequency, so we have more time to handle the data when we get it.

src/rtde/rtde_parser.cpp
```diff
@@ -0,0 +1,174 @@
...
+ */
+#include "ur_client_library/rtde/rtde_parser.h"
+#include "ur_client_library/rtde/package_header.h"
+#include "ur_client_library/rtde/rtde_package.h"
+
+namespace urcl
+{
+namespace rtde_interface
+{
+
```
urrsk, line 30: - [ ] Please add a not and a warning that this is deprecated and does not have a good realtime performance

src/rtde/rtde_parser.cpp
```diff
@@ -0,0 +1,174 @@
...
+  PackageType type = getPackageTypeFromHeader(bp);
+
+  switch (type)
+  {
+    case PackageType::RTDE_DATA_PACKAGE:
+    {
+      if (result == nullptr || result->getType() != PackageType::RTDE_DATA_PACKAGE)
+      {
+        result = std::make_unique<DataPackage>(recipe_, protocol_version_);
+        URCL_LOG_WARN("The passed result pointer is empty or does not contain a DataPackage. A new DataPackage will "
```
urrsk: - [ ] Let's us ensure that this warning only applies when there is something unexpected happening

src/rtde/rtde_client.cpp
urfeex: This triggers only if getting a new data package failed which could be one of the cases:

- producer isn't running
- stream is closed (disconnected)
- reconnection is active

That frequency is merely a slowing down bombarding the producer with queries when it's not able to deliver data, anyway. Once everything is up and running again, there is no sleep, but the pace is actually given by the data coming in.

src/rtde/rtde_writer.cpp
```diff
@@ -161,7 +217,10 @@ bool RTDEWriter::sendConfigurableDigitalOutput(uint8_t output_pin, bool value)
     return false;
   }
 
-  std::lock_guard<std::mutex> guard(package_mutex_);
+  URCL_LOG_INFO("Write thread started.");
```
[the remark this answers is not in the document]
urrsk: - [ ] Move the log_info

src/rtde/rtde_writer.cpp
```diff
@@ -62,28 +109,49 @@ void RTDEWriter::run()
...
     writer_thread_.join();
   }
 }
 
+bool RTDEWriter::sendPackage(const DataPackage& package)
+{
+  *current_store_buffer_ = package;
+  markStorageToBeSent();
+  return true;
+}
```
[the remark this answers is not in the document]
urrsk: - [ ] Verify this

src/rtde/rtde_writer.cpp
```diff
@@ -27,32 +27,79 @@
...
+#include <mutex>
+#include "ur_client_library/log.h"
 
 namespace urcl
 {
 namespace rtde_interface
 {
+std::vector<std::string> RTDEWriter::g_preallocated_input_bit_register_keys;
+std::vector<std::string> RTDEWriter::g_preallocated_input_int_register_keys;
+std::vector<std::string> RTDEWriter::g_preallocated_input_double_register_keys;
```
urrsk: Is it okay that these are shared across objects (stativ variables)

urfeex: Yes, that's the intention. I'll update this together with https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/419#discussion_r2768885745 and mark them const.

include/ur_client_library/comm/producer.h
```diff
@@ -47,6 +47,57 @@ class URProducer : public IProducer<T>
...
+      if (stream_.read(buf, sizeof(buf), read))
+      {
+        // reset sleep amount
+        timeout_ = std::chrono::seconds(1);
+        BinParser bp(buf, read);
+        return parser_.parse(bp, product);
+      }
+
+      if (!running_)
+        return false;
```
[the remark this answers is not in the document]
urfeex: From my point of view that is fine and intended.

src/rtde/rtde_parser.cpp
urfeex: With the latest changes I could not encounter this during normal operation. I've added a more descriptive output in that case so if it ever happened, we know a bit better what happened.

src/rtde/rtde_parser.cpp
```diff
@@ -106,10 +106,20 @@ bool RTDEParser::parse(comm::BinParser& bp, std::unique_ptr<RTDEPackage>& result
...
+                        "would be sent.");
+        }
+        else
+        {
+          URCL_LOG_WARN("Passed a pre-allocated RTDE package of type %u while a DataPackage was received. A new "
+                        "DataPackage will have to be allocated. Please pass a pre-allocated DataPackage if you expect "
+                        "a DataPackage would be sent.",
+                        result->getType());
+        }
         result = std::make_unique<DataPackage>(recipe_, protocol_version_);
```
urrsk, line 122: - [ ] Consider make one if elseif case instead of an if else in an if

urfeex: The `result = std::make_unique<DataPackage>(recipe_, protocol_version_);` is common for both cases, which is why I catch those cases in the outer if. Depending on what's the cause for a not correctly allocated data package the output is different, hence the inner if.

Merging both layers into a single `if...else if` would require duplicating the important line potentially increasing the error missing one case when updating the code in the future. My opinion on this isn't that strong, if you think that the following structure is better, we can go with that:

```c++
      if (result == nullptr)
      {
        URCL_LOG_WARN("The passed result pointer is empty. A new DataPackage will "
                      "have to be allocated. Please pass a pre-allocated DataPackage if you expect a DataPackage "
                      "would be sent.");
        result = std::make_unique<DataPackage>(recipe_, protocol_version_);
      }
      else if (result->getType() != PackageType::RTDE_DATA_PACKAGE)
      {
        URCL_LOG_WARN("Passed a pre-allocated RTDE package of type %u while a DataPackage was received. A new "
                      "DataPackage will have to be allocated. Please pass a pre-allocated DataPackage if you expect "
                      "a DataPackage would be sent.",
                      result->getType());
        result = std::make_unique<DataPackage>(recipe_, protocol_version_);
      }
```

urrsk (approved):

src/rtde/rtde_client.cpp
```diff
@@ -852,9 +943,76 @@ void RTDEClient::reconnectCallback()
   {
     reconnecting_thread_.join();
   }
-  reconnecting_ = true;
   reconnecting_thread_ = std::thread(&RTDEClient::reconnect, this);
 }
 
+void RTDEClient::startBackgroundRead()
```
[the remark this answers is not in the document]
urfeex: That's right. I made the change, since we had a test triggering `reconnect` directly, which would have never set that flag, then. But since we removed that test by now, we can move setting the flag back to the callback.

urrsk (approved):

Issue #423, opened 2026-01-30
After running 1 trajectory, no more trajectories can be ran


PabloDArandaR:
### Affected Client Library version(s)

2.6.1

### What combination of platform is the Client Library running on.

Linux

### How did you install the Client Library

Build the library from source

### Which robot platform is the library connected to.

UR E-series robot

### Robot SW / URSim version(s)

5.12

### How are you using the Client Library

Headless without using the teach pendant

### Issue details

# Summary
After using a moveL call with the instruction executor, no additional trajectories can be ran, although starting my program again does allow us to run another trajectory

## Issue details
When I try to run the another trajectory, I get the following log error:
```
[1769787959.639515] ERROR /home/pablo/Projects/externals-linux/_build/universal_robots_client_library/src/ur/ur_driver.cpp 478: Script command interface is not running. Unable to end Force mode.
[1769787959.639836] ERROR /home/pablo/Projects/externals-linux/_build/universal_robots_client_library/src/ur/instruction_executor.cpp 64: Cannot send trajectory control command. No client connected?
```

After I run the first trajectory, I get the following messages:
```
[1769788297.479357] INFO /home/pablo/Projects/externals-linux/_build/universal_robots_client_library/src/control/reverse_interface.cpp 238: Connection to reverse interface dropped.
[1769788297.479420] INFO /home/pablo/Projects/plugins/RoboticsPlugin/Source/UR/URRunner.cpp 114: Program state changed: Stopped
[1769788297.479373] INFO /home/pablo/Projects/externals-linux/_build/universal_robots_client_library/src/ur/instruction_executor.cpp 49: Trajectory disconnect
```

Additionally, I checked that the runtime states passes to be STOPPED after the first trajectory.

## Steps to Reproduce
Try to run a trajectory, and then another one.

## Expected Behavior
To be able to run additional trajecotry

# Workaround Suggestion

They way I managed to sort this out temporarily, is by sending again the RobotProgram between executions:
```
			try
			{
				m_robotHandle->sendRobotProgram();

				// Wait for program to start with timeout
				std::unique_lock<std::mutex> lk(m_programMutex);
				if (!m_programState.wait_for(lk, std::chrono::seconds(3), [this] { return m_programRunning; }))
				{
					LOG_ERROR("Failed to restart URScript program within timeout");
					return {std::nullopt};
				}
				lk.unlock();

				// Send keepalive and wait a bit for connection to stabilize
[...]

PR #424, opened 2026-02-02
Fix instruction executor idle

urfeex:
When using the InstructionExecutor, sending the first trajectory puts the external_control program in `MODE_FORWARD`. We then send keepalive signals every 100ms with a timeout of 200 ms.

However, when the trajectory has finished, nothing is sent anymore, hence reading the keepalive signal times out and program execution is stopped. This is most likely the cause of #423.

This PR explicitly sets the external_control program into `MODE_IDLE` once a trajectory has finished. That allows waiting for a new package on the reverse interface indefinitely. Thus, users can spend as much time between trajectory calls as they like without having to manually send keepalive signals of some sort.


PabloDArandaR:
Everything is good from my side, I did have to reduce the timeout of my monitoring function, but besides that not "hack" is now needed

PR #435, opened 2026-02-12
Add more RobotAPI commands to PolyScope X Dashboard Client

urfeex:
This adds the endpoints as added in [PolyScope 10.12.0](https://www.universal-robots.com/articles/ur/release-notes/release-note-software-version-1012x/)

### DIFF
--- a/tests/test_dashboard_client_x.cpp
+++ b/tests/test_dashboard_client_x.cpp
@@ -382,6 +382,10 @@ TEST_F(DashboardClientTestX, upload_program_from_file)
       URCL_LOG_INFO("status code: %d", std::get<int>(response.data["status_code"]));
       ASSERT_EQ(std::get<int>(response.data["status_code"]), error_code_exists);
     }
+
+    response = dashboard_client_->commandUploadProgram("non_existent_file.urpx");
+    ASSERT_FALSE(response.ok);
+    ASSERT_EQ(response.message, "URPX File not found: non_existent_file.urpx");
   }
 }
 


include/ur_client_library/ur/dashboard_client.h
```diff
@@ -748,8 +792,22 @@ class DashboardClient
...
+   * If the dashboard client implementation is not able to query the version from the robot
+   * automatically, this function can be used to set it manually.
+   *
+   * \param version The version string as returned by the robot
+   */
+  void setPolyscopeVersion(const VersionInformation& version)
+  {
+    polyscope_version_ = version;
+  }
+
```
urrsk: When is this necessary?

src/ur/dashboard_client_implementation_g5.cpp
```diff
@@ -1186,6 +1186,27 @@ DashboardResponse DashboardClientImplG5::commandSaveLog()
...
+{
+  throw NotImplementedException("commandUpdateProgram is not implemented for DashboardClientImplG5.");
+}
+
+DashboardResponse DashboardClientImplG5::commandDownloadProgram(const std::string& program_file_name,
+                                                                const std::string& destination_path)
+{
+  throw NotImplementedException("commandDownloadProgram is not implemented for DashboardClientImplG5.");
+}
+
```
urrsk, line 1213: I will suggest we write " command*Program is supported from PolyScopeX 10.12.0" pointing at the robot controller version and not the URCL

urfeex: Sounds like a good idea.

include/ur_client_library/ur/dashboard_client.h
urfeex: The dashboard client itself doesn't necessarily know its version (at least on PolyScope X). With this, we can query the version elsewhere (e.g. the primary interface) and let the Robot API client know which version it is running on.

urrsk: would the version of the Robot API be enough?

urfeex: That's a good suggestion! I've changed the code accordingly.

PR #440, opened 2026-02-17
If no new commands are received, halt

urfeex:
Currently, we only exit the main loop when no new command is received anymore (already taking any configured receive timeout into account).

However, if we don't receive a new command anymore, that should always be considered an unintentional interruption of the control flow. In that case, we should stop program execution, as interrupting the communication was an unexpected error event.

If interrupting the communication in order to continue with the rest of the program was intentional, users will send a stop command. This will end the main control loop at another code branch leading to a clean shutdown of the external_control part in order to be able to continue with the rest of the program.

This should be closing #438 but in order to make sense with the ROS drivers, we would need https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1678 and potentially https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/767 before merging this.

---

> [!NOTE]
> **Medium Risk**
> Changes robot program termination behavior on communication loss and adjusts core TCP server concurrency/shutdown logic; regressions could impact control flow or socket handling under load.
> 
> **Overview**
> Updates `resources/external_control.urscript` so a reverse-socket receive timeout is treated as a fatal control interruption: it now performs `stopj(STOPJ_ACCELERATION)` and `halt` instead of continuing/cleanly stopping.
> 
> Hardens `comm::TCPServer` with mutexes around callback registration/invocation and client list mutation (including shutdown), tweaks disconnect/destructor logging to INFO, and ensures shutdown clears client state and invalidates `listen_fd_`.
> 
> Adds an integration test (`test_external_control_program.cpp`) for the new timeout behavior and refactors multiple socket/pipeline/stream/server tests to use a shared `TestableTcpServer` helper (plus CMake wiring) to reduce callback boilerplate and avoid teardown races.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 636d29b30240c5fc07baf82a77fc1e0320a19a04. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/resources/external_control.urscript
+++ b/resources/external_control.urscript
@@ -1003,7 +1003,8 @@ while control_mode > MODE_STOPPED:
     end
   else:
     textmsg("Socket timed out waiting for command on reverse_socket. The script will exit now.")
-    control_mode = MODE_STOPPED
+    stopj(STOPJ_ACCELERATION)
+    halt
   end
   exit_critical
 end
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -82,6 +82,22 @@ if (INTEGRATION_TESTS)
                   TEST_SUFFIX _headless
   )
 
+  # ExternalControlProgram tests
+  add_executable(external_control_program_tests_urcap test_external_control_program.cpp)
+  target_link_libraries(external_control_program_tests_urcap PRIVATE ur_client_library::urcl GTest::gtest_main)
+  gtest_add_tests(TARGET      external_control_program_tests_urcap
+                  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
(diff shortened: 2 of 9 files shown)


urrsk (approved):

urrsk (changes requested):
We should add a test covering this.

urfeex:
> We should add a test covering this.

Sounds good, I also have an idea for that in mind. It's a bit more complicated, as we need to generate a PolyScope program and require the EC URCap to run that, but it should be possible to do. I just don't know when I will have the time to spend on this.

urrsk (dismissed):

urrsk (approved):

PR #480, opened 2026-04-08
Add parsing of safety mode messages to the primary interface

URJala:
This is necessary to implement script execution feedback in the the primary interface.

### DIFF
--- a/src/primary/robot_message/safety_mode_message.cpp
+++ b/src/primary/robot_message/safety_mode_message.cpp
@@ -84,7 +84,7 @@ bool SafetyModeMessage::consumeWith(AbstractPrimaryConsumer& consumer)
 std::string SafetyModeMessage::toString() const
 {
   std::stringstream ss;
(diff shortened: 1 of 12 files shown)


src/primary/robot_message/safety_mode_message.cpp
```diff
@@ -0,0 +1,112 @@
...
+  }
+  return true;
+}
+
+bool SafetyModeMessage::consumeWith(AbstractPrimaryConsumer& consumer)
+{
+  return consumer.consume(*this);
+}
+
+std::string SafetyModeMessage::toString() const
```
urfeex, line 84: I think, it would make sense to add a test for this, especially because of the different formatting depending on the report data type. You could use the same binary data for the `report_data` and assert that it gets formatted correctly depending on the `report_data_type`.

urfeex (changes requested):

urfeex (approved):

PR #482, opened 2026-04-10
Fix traj point time precision

urfeex:
Precision of trajectory point duration was using floor-rounding on millisecond precision. Thus, e.g. `0.019998 sec` would be transferred as `19 ms`.

This PR changes the following:

- Change time precision for time from milliseconds to microseconds. This effectively makes the longest duration for a trajectory point ~36 minutes, which in my opinion is fair.
- Use nearest integer round instead of floor round if time with more precision than 1 microsecond is given
- Use Micrometer precision for the blend radius. Probably not that much relevant, but it is using the same multiplier constant.

Backwards compatibility should be given, as long as users haven't hardcoded the constant's value. The change on the wire protocol is injected by the application, so even a custom script code would get the updated constant injected correctly decoding the changed wire data.

### DIFF
--- a/doc/architecture/trajectory_point_interface.rst
+++ b/doc/architecture/trajectory_point_interface.rst
@@ -70,3 +70,8 @@ where
 
 - ``MULT_JOINTSTATE``: 1000000
 - ``MULT_TIME``: 1000000
+
+.. note::
+   With ``MULT_TIME`` being 1000000, the maximum duration that can be sent is 2147 seconds, while
+   precision is cut off at 1 microsecond. (The same applies to the blend radius, respectively being
+   max 2147 m and 1 μm precision.)
--- a/doc/architecture/trajectory_point_interface.rst
+++ b/doc/architecture/trajectory_point_interface.rst
@@ -59,14 +59,13 @@ representations in 21 datafields. The data fields have the following meaning:
           - 3: MOVEC
           - 51: SPLINE)
 
-   19     trajectory point time (in seconds, multiplied by ``MULT_TIME``)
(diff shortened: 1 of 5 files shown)


urrsk (changes requested):
- [x] Please make sure it rejects a trajectory the exceed the time limit and provide relevant feedback to the user why the trajectory is rejected

gavanderhoorn:
as I've worked on (other) drivers:

> This effectively makes the longest duration for a trajectory point ~36 minutes, which in my opinion is fair.

(goals with) trajectories long than the supported maximum should be rejected explicitly and the driver should clearly explain why.

I've had to debug applications where users were trying to 'plan' a whole day worth of trajectories/movements ahead of time, and sending it as a single trajectory ("to avoid startup delays as much as possible").

Just a data point that very / extremely long trajectories are not necessarily very rare.

---

Edit: if this is not about trajectories, but really trajectory *points* (and it's about encoding the duration of the segment *towards* the next point) then perhaps `/ignore`.

urfeex:
@gavanderhoorn yes, this is only about segment durations, but as @urrsk pointed out, catching this case and trigger some sort of error handling seems to be the right thing to do, even if it is a rare edge case.

urfeex:
I've added a check for `durations[µs] > maxint32` and reject waypoints with those. Also added a test for that. I would prefer throwing an exception, so users wouldn't have to scroll through the logs to find the reason for rejection, but this is concise with the other errors that we raise. Raising exceptions for error handling probably has to be moved to an API-breaking change somewhere in the future.

urrsk (approved):

PR #489, opened 2026-04-21
Fix: primary_configuration_timeout as configurable field in UrDriverConfiguration

litziacarla:
This PR adds primary_configuration_timeout as a configurable field in UrDriverConfiguration, replacing the previously hardcoded 1000 ms timeout used when waiting for primary client configuration data during driver initialization, since in some network environments, the hard coded value is insufficient and causes initialization failures.

This change is a prerequisite for [Universal_Robots_ROS2_Driver#primary_configuration_timeout as hardware interface parameter], which wires this parameter through the ROS 2 hardware interface and URDF/xacro description.

---

> [!NOTE]
> **Low Risk**
> Low risk: only replaces a hardcoded 1s initialization timeout with a config-driven value, changing startup behavior but not runtime control logic.
> 
> **Overview**
> Makes the primary-client configuration wait timeout configurable via a new `UrDriverConfiguration::primary_configuration_timeout` field (defaulting to 1s).
> 
> Driver initialization now uses this value instead of a hardcoded 1000ms when waiting for `primary_client_->getConfigurationData()`, improving robustness in slower network environments and producing the same `TimeoutException` with the configured duration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02f84afd77dbb455ff4e7ac5d962c76c857d194c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/include/ur_client_library/ur/ur_driver.h
+++ b/include/ur_client_library/ur/ur_driver.h
@@ -136,6 +136,11 @@ struct UrDriverConfiguration
    */
   std::chrono::milliseconds rtde_initialization_timeout = std::chrono::seconds(5);
 
+  /*!
+   * \brief Timeout for waiting for primary client configuration data.
+   */
+  std::chrono::milliseconds primary_configuration_timeout = std::chrono::seconds(1);
+
   bool non_blocking_read = false;
 
   // TODO: Remove on 2027-05
--- a/include/ur_client_library/ur/ur_driver.h
+++ b/include/ur_client_library/ur/ur_driver.h
@@ -136,11 +136,6 @@ struct UrDriverConfiguration
    */
   std::chrono::milliseconds rtde_initialization_timeout = std::chrono::seconds(5);
 
-  /*!
-   * \brief Timeout for waiting for primary client configuration data.
-   */
-  std::chrono::milliseconds primary_configuration_timeout = std::chrono::seconds(1);
-
   bool non_blocking_read = false;
 
   // TODO: Remove on 2027-05
--- a/src/ur/ur_driver.cpp
+++ b/src/ur/ur_driver.cpp
@@ -98,7 +98,7 @@ void UrDriver::init(const UrDriverConfiguration& config)
 
   startPrimaryClientCommunication();
 
-  std::chrono::milliseconds timeout(1000);
+  std::chrono::milliseconds timeout(config.primary_configuration_timeout);
   try
   {
     waitFor([this]() { return primary_client_->getConfigurationData() != nullptr; }, timeout);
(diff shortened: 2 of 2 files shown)


urfeex:
I don't see a reason not to add this, but I would be curious in what kind of network environment this is an issue. In my local tests I never had more than 100ms between calling `startPrimaryClientCommunication` and having received a ConfigurationData package. Not more than 150ms when going through a vpn with a ping of ~22 ms to the robot.

litziacarla:
In our setup we launch many nodes simultaneously (MoveIt, camera, RViz, mesh processing and others...), and the CPU load at startup is enough to delay the first primary client configuration packet beyond the hardcoded 1 second timeout. I measured a delay of ~1.3s in my case.

urfeex (commented):
Looking back at #381 (where this wait block was added) it seems that we don't really need this right now. It was added to be able to safely get the robot model which was needed for selecting the right parameter set for the PD controller.

However, that feature has been removed from #381 and thus, this is no longer needed. However, it will be needed for #400 (which is basically the same feature, but factored out to its own PR).

@urrsk what do you think? Will we need that in the final version of #400? I guess yes, as we plan to have individual parameter sets for each robot type, right? Thus, it makes sense to wait for a Configuration data package in the beginning.

However, if the timeout can be hit in systems with heavy load, we might want to increase the default value. It is not really a timeout in terms of "Something is going too slow here", it's more a "Wait until you get this. We'll add a timeout here, just in case that we ever get to this point without a real primary interface being on the other side and we would like to avoid waiting forever on something that will never happen".

litziacarla:
[#1772](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1772)

urfeex:
@litziacarla I think, I would rather like to increase the timeout to something like 10 seconds rather than adding yet another parameter. As written above, it is more of a "work or die" feature and we basically only want to catch the unlikely case that the driver connects to something that is not a UR primary interface and hangs there forever. Would you be fine with that?

litziacarla:
> @litziacarla I think, I would rather like to increase the timeout to something like 10 seconds rather than adding yet another parameter. As written above, it is more of a "work or die" feature and we basically only want to catch the unlikely case that the driver connects to something that is not a UR primary interface and hangs there forever. Would you be fine with that?

Sure, 10 seconds should be more than enough. Thanks for addressing this issue!

urfeex:
Would you like to  change this PR or should I close this one and open a new one?

litziacarla:
> Would you like to change this PR or should I close this one and open a new one?

I've just updated this PR with a hardcoded 10-second timeout

urfeex (approved):

PR #511, opened 2026-05-20
Refactor TCPSocket::setup Timeout and Error Handling

srvald:
### Description
This PR refactors the `TCPSocket::setup()` method to resolve the blocking behavior during network failures. 

The implementation now introduces an asynchronous approach to enforce a connection timeout and capture socket errors.

### Step-by-Step Modifications

**1. Signature Update (Timeout Parameter)**
* Added a new `timeout` parameter (`std::chrono::milliseconds`) to the `setup()` function. This makes it possible to define exactly how long to wait for a connection before giving up. Otherwise, the code blocks indefinitely, keeping the thread hanging even if the number of max tries is set to 1. The default value has been set to 500 ms. 

**2. Error Tracking (`std::error_code`)**
* Introduced a `std::error_code socket_error` variable. 
* Now, `getLastSocketErrorCode()` is called immediately after any failure to store the exact system reason (e.g., `ECONNREFUSED`, `EHOSTUNREACH`) so it can be reported later.

**3. Non-Blocking Connection Logic**
* **State Change:** The socket is temporarily set to non-blocking mode using `fcntl()` and `O_NONBLOCK`.
* **Async Connect:** `::connect()` is called. If it returns `EINPROGRESS`, the code now hands control over to `::select()`.
* **Timeout Enforcement:** `select()` waits for the socket to become writable, strictly bounded by the newly provided `timeout` parameter.
* **Error Verification:** If `select()` indicates readiness, `::getsockopt(..., SO_ERROR, ...)` is used to verify if the connection actually succeeded or if it failed in the background.
* **State Restoration:** The socket is restored to its original blocking state before proceeding.

**4. Proper Resource Cleanup on Failure**
* If `socket_fd_` is created but the connection attempt fails, it is now properly closed.
* Added `::ur_close(socket_fd_)` and reset to `INVALID_SOCKET` inside the failure branches within the `for` loop to properly release resources. Also added error capturing if `::socket()` creation fails directly.

**5. Improved Logging & Debugging**
* Replaced the generic and hardcoded log message (`"Please check that the robot is booted and reachable..."`) with the exact system string translated by `socket_error.message()`. 
* Logs will now report `"Reason: Connection refused"`, `"Reason: Operation not permitted"`, or `"Reason: Connection timed out"`, improving troubleshooting capabilities for end users.

### Testing Performed

- **Connection Refused:** Pointed to an active IP with a closed port. Verified the code bypasses `select()` and fails immediately.

image

- **Connection Timed Out:** Simulated a network blackhole (e.g., using a dead IP). Verified the code waits exactly for the `timeout` duration before logging a timeout error.

image

- **Host Unreachable:** Pointed to an invalid subnet. Verified immediate routing error capture.

image

---

> [!NOTE]
> **Medium Risk**
> Changes core TCP connection establishment logic (non-blocking connect + `select()`/`getsockopt()`), which can affect connectivity behavior across platforms and edge cases like fd limits/timeouts.
> 
> **Overview**
> `TCPSocket::setup()` now accepts an additional `timeout` (default `500ms`) and uses a non-blocking connect with `select()` to enforce connection timeouts instead of potentially blocking indefinitely.
> 
> Connection attempts now capture and surface the underlying `std::error_code` reason in retry logs, add extra failure-path cleanup (close/reset fd), and include a non-Windows guard for `FD_SETSIZE` overflow before using `select()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b46158a2b92273be8bc6d3f667c90cf04e4f30c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/include/ur_client_library/comm/tcp_socket.h
+++ b/include/ur_client_library/comm/tcp_socket.h
@@ -56,11 +56,6 @@ private:
   void setupOptions();
 
 protected:
-  static bool open(socket_t socket_fd, struct sockaddr* address, size_t address_len)
-  {
-    return ::connect(socket_fd, address, static_cast<socklen_t>(address_len)) == 0;
-  }
-
   bool setup(const std::string& host, const int port, const size_t max_num_tries = 0,
              const std::chrono::milliseconds reconnection_time = DEFAULT_RECONNECTION_TIME,
              const std::chrono::milliseconds timeout = DEFAULT_CONNECTION_TIME);
--- a/include/ur_client_library/comm/tcp_socket.h
+++ b/include/ur_client_library/comm/tcp_socket.h
@@ -62,12 +62,14 @@ protected:
   }
 
   bool setup(const std::string& host, const int port, const size_t max_num_tries = 0,
-             const std::chrono::milliseconds reconnection_time = DEFAULT_RECONNECTION_TIME);
+             const std::chrono::milliseconds reconnection_time = DEFAULT_RECONNECTION_TIME,
+             const std::chrono::milliseconds timeout = DEFAULT_CONNECTION_TIME);
 
   std::unique_ptr<timeval> recv_timeout_;
 
 public:
   static constexpr std::chrono::milliseconds DEFAULT_RECONNECTION_TIME{ 10000 };
+  static constexpr std::chrono::milliseconds DEFAULT_CONNECTION_TIME{ 500 };
   /*!
    * \brief Creates a TCPSocket object
    */
--- a/src/comm/tcp_socket.cpp
+++ b/src/comm/tcp_socket.cpp
@@ -171,31 +171,33 @@ bool TCPSocket::setup(const std::string& host, const int port, const size_t max_
         tv.tv_sec = static_cast<long>(timeout_ms / 1000);
         tv.tv_usec = static_cast<long>((timeout_ms % 1000) * 1000);
 
+        socket_t local_fd = socket_fd_.load();
+
         fd_set write_fds;
         FD_ZERO(&write_fds);
-        FD_SET(socket_fd_, &write_fds);
+        FD_SET(local_fd, &write_fds);
 
         fd_set except_fds;
         FD_ZERO(&except_fds);
-        FD_SET(socket_fd_, &except_fds);
+        FD_SET(local_fd, &except_fds);
 
-        int select_res = ::select(static_cast<int>(socket_fd_ + 1), nullptr, &write_fds, &except_fds, &tv);
+        int select_res = ::select(static_cast<int>(local_fd + 1), nullptr, &write_fds, &except_fds, &tv);
 
         if (select_res > 0)
         {
           int so_error = 0;
 #ifndef _WIN32
           socklen_t len = sizeof(so_error);
-          int opt_res = ::getsockopt(socket_fd_, SOL_SOCKET, SO_ERROR, &so_error, &len);
+          int opt_res = ::getsockopt(local_fd, SOL_SOCKET, SO_ERROR, &so_error, &len);
 #else
           int len = sizeof(so_error);
-          int opt_res = ::getsockopt(socket_fd_, SOL_SOCKET, SO_ERROR, reinterpret_cast<char*>(&so_error), &len);
+          int opt_res = ::getsockopt(local_fd, SOL_SOCKET, SO_ERROR, reinterpret_cast<char*>(&so_error), &len);
 #endif
           if (opt_res != 0)
           {
             socket_error = getLastSocketErrorCode();
           }
-          else if (so_error == 0 && FD_ISSET(socket_fd_, &write_fds))
+          else if (so_error == 0 && FD_ISSET(local_fd, &write_fds))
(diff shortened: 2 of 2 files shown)


urfeex:
To be honest, I'm not sure whether this actually improves things a lot. Am I correct, that you want to solve https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/838?

From my understanding, that issue isn't about the blocking connect itself. For example, you can start the primary pipeline example from the client library with a wrong IP / an IP address that doesn't host a primary interface and just close it using CTRL-C.

Thus, I would expect that https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/838 is more a problem of signal forwarding inside the ROS node.

This PR seems to complicate the socket connection quite a lot and I'd like to challenge the requirement for actually doing this.

srvald:
Hi @urfeex. Well... this PR doesn't fully resolve [#838] https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/838) on its own. 

Regarding the `primary_pipeline` example you mentioned, the `::connect()` call does block the thread, but since it's a standard C++ application, when CTRL-C is hit, the OS terminates the process immediately, but  it is able to do it because is a standard application, isn't it? 

With this PR implemented, the output on the teminal when using a wrong IP for that example will be:

```bash
~/ws_rol/ur_c/Universal_Robots_Client_Library/b/examples tcp_socket_timeout > ./primary_pipeline_example 1.2.3.4 
[1780382828.827426] ERROR /home/mirserv/ws_rolling/ur_client_library/Universal_Robots_Client_Library/src/comm/tcp_socket.cpp 270: Failed to connect to robot on IP 1.2.3.4:30001. Reason: Connection timed out. Retrying in 10 seconds. 
[1780382839.328329] ERROR /home/mirserv/ws_rolling/ur_client_library/Universal_Robots_Client_Library/src/comm/tcp_socket.cpp 270: Failed to connect to robot on IP 1.2.3.4:30001. Reason: Connection timed out. Retrying in 10 seconds. 
[1780382849.829394] ERROR /home/mirserv/ws_rolling/ur_client_library/Universal_Robots_Client_Library/src/comm/tcp_socket.cpp 270: Failed to connect to robot on IP 1.2.3.4:30001. Reason: Connection timed out. Retrying in 10 seconds. 
[1780382860.330338] ERROR /home/mirserv/ws_rolling/ur_client_library/Universal_Robots_Client_Library/src/comm/tcp_socket.cpp 270: Failed to connect to robot on IP 1.2.3.4:30001. Reason: Connection timed out. Retrying in 10 seconds. 
```

Without the PR, it will be blocked without showing any output:

```bash
~/ws_rol/ur_c/Universal_Robots_Client_Library/b/examples master > ./primary_pipeline_example 1.2.3.4 
^C
~/ws_rol/ur_c/Universal_Robots_Client_Library/b/examples master > 
```

In a ROS 2 node like `urscript_interface`, it is different. ROS 2 captures the CTRL+C signal to do a clean shutdown, meaning the OS doesn't just kill the process anymore, as done in the C++ application, because ROS 2 catches it and flips its internal shutdown flag, but it cannot interrupt the socket block. So, the node just hangs there.

Currently, when `urscript_interface` is used with a wrong IP, it gets trapped in this chain:

1. `m_secondary_stream->connect()` is called in the constructor.

https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/417f7323bea6472fa52077aea0428c3b7b273a5a/ur_robot_driver/src/urscript_interface.cpp#L72

2. It enters `TCPSocket::setup` and hits the `while (!connected)` loop.

https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/627cf205cec22a727b73e7c317d93fc088f68ce4/src/comm/tcp_socket.cpp#L75-L76

https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/627cf205cec22a727b73e7c317d93fc088f68ce4/src/comm/tcp_socket.cpp#L107

3. It calls the OS `::connect()`, which has no timeout.

https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/627cf205cec22a727b73e7c317d93fc088f68ce4/include/ur_client_library/comm/tcp_socket.h#L61

Because of this, the thread is blocked. This makes the `max_num_tries` parameter completely useless, it gets stuck on the very first try and freezes the node, preventing any CTRL+C response.

This PR introduces a timeout using `select()`. Now, if the IP is unreachable, the attempt times out quickly (e.g., 500ms). This allows `max_num_tries` to actually work as intended, iterate, and eventually return that it couldn't connect, freeing the thread.

To fix this at the node level, I actually tested moving `m_secondary_stream->connect()` into a ROS Timer with `max_num_tries` set to 1, and it worked perfectly (the main node didn't block and CTRL+C worked instantly). However, if I recall correctly, we previously discussed holding off on that node-level change because the upcoming primary client refactor might handle it better.

Let me know if this makes sense or if you think I should take a different approach.

PR #518, opened 2026-06-11
[DashboardClientX] Apply httplib timeouts; wire setReceiveTimeout()

Vinicius-T-Robotics:
cpp-httplib's default read_timeout is 300 seconds. Because the PolyScope X DashboardClientImplX constructor never overrides any timeout, every dashboard call (commandPowerOff, commandStop, commandPlay, etc.) blocks for the full 300 s when the controller becomes unresponsive — e.g. a network partition or paused container in tests. Symptoms include thread-pool exhaustion in callers that dispatch dashboard commands from a bounded executor pool.

Two changes:

1. Apply explicit set_connection_timeout(5s), set_read_timeout(1s), and set_write_timeout(5s) in the constructor. The 1 s read default matches the contract already documented on this impl: getConfiguredReceiveTimeout() returns {tv_sec = 1, tv_usec = 0}.

2. Wire setReceiveTimeout(const timeval&) into the underlying httplib::Client instead of being a [[maybe_unused]] no-op. This makes the base-class API work as documented on the X path; callers can now tune the read timeout per their needs.

Both changes are backward-compatible: there is no public API change, and the new defaults align with what the impl already advertised via getConfiguredReceiveTimeout().

---

> [!NOTE]
> **Medium Risk**
> Changes how long all X dashboard HTTP operations block on failure and alters default timeouts versus the previous effective 300 s read behavior; power-on now honors its timeout parameter on the HTTP layer.
> 
> **Overview**
> PolyScope X dashboard HTTP calls no longer inherit cpp-httplib’s **300 s** read default. `DashboardClientImplX` sets **5 s** connection timeout and applies stored **read/write** timeouts (default **10 s** each) when the `httplib::Client` is created.
> 
> **Receive/send timeout API** is implemented on the X path: `setReceiveTimeout` / `setSendTimeout` update the client and `getConfiguredReceiveTimeout` / `getConfiguredSendTimeout` return the stored `timeval` values. The base `DashboardClientImpl` gains optional `getConfiguredSendTimeout` / `setSendTimeout` hooks (1 s default on non-overriding impls).
> 
> `commandPowerOn` temporarily raises the read timeout to the caller’s duration for the power-on PUT, then restores the previous setting (including on exceptions). Tests cover timeout getters/setters, sub-second precision, and a tiny receive timeout failing `connect`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26a3ec550c2591c613f70898bf9b873436cc30da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Vinicius-T-Robotics:
Hi @urfeex — would you be willing to take a look at this when you have a moment? Small fix for the dashboard 300 s hang on PolyScope X. Thanks!

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -49,6 +50,22 @@ DashboardClientImplX::DashboardClientImplX(const std::string& host) : DashboardC
...
   // work, we'll have to follow redirects, which is not the default for httplib.
   cli_->set_follow_location(true);
+
+  // cpp-httplib's default read_timeout is 300 seconds. Applied unchanged, this makes any
+  // dashboard call hang for 5 minutes when the controller becomes unresponsive (e.g. network
+  // partition, paused container in tests). Align with the contract documented on the base
+  // class: getConfiguredReceiveTimeout() returns {tv_sec = 1, tv_usec = 0} on this impl.
+  cli_->set_connection_timeout(std::chrono::seconds(5));
+  cli_->set_read_timeout(std::chrono::seconds(1));
+  cli_->set_write_timeout(std::chrono::seconds(5));
```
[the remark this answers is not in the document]
Vinicius-T-Robotics: fixed in the follow-up commit.

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -68,19 +95,23 @@ bool DashboardClientImplX::connect([[maybe_unused]] const size_t max_num_tries,
...
+        robot_api_version_ = VersionInformation::fromString(json_data["info"]["version"]);
+        URCL_LOG_DEBUG("Connected to Robot API version: %s", robot_api_version_.toString().c_str());
+        result = true;
+      }
     }
   }
-  return false;
+
+  setReceiveTimeout(configured_tv);
+  return result;
```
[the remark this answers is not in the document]
Vinicius-T-Robotics: Fixed in follow up commit

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -68,19 +95,23 @@ bool DashboardClientImplX::connect([[maybe_unused]] const size_t max_num_tries,
...
-    auto json_data = json::parse(db_res.message);
-    if (db_res.ok && json_data.contains("info") && json_data["info"].contains("version") &&
-        json_data["info"]["version"].is_string())
+    else
     {
-      robot_api_version_ = VersionInformation::fromString(json_data["info"]["version"]);
-      URCL_LOG_DEBUG("Connected to Robot API version: %s", robot_api_version_.toString().c_str());
-      return true;
+      auto db_res = handleHttpResult(res, false);
+      auto json_data = json::parse(db_res.message);
```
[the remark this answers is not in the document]
Vinicius-T-Robotics: Fixed

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -49,6 +50,22 @@ DashboardClientImplX::DashboardClientImplX(const std::string& host) : DashboardC
   // Some targets have changed between versions redirecting to the correct endpoint. For this to
   // work, we'll have to follow redirects, which is not the default for httplib.
   cli_->set_follow_location(true);
+
+  // cpp-httplib's default read_timeout is 300 seconds. Applied unchanged, this makes any
+  // dashboard call hang for 5 minutes when the controller becomes unresponsive (e.g. network
+  // partition, paused container in tests). Align with the contract documented on the base
+  // class: getConfiguredReceiveTimeout() returns {tv_sec = 1, tv_usec = 0} on this impl.
+  cli_->set_connection_timeout(std::chrono::seconds(5));
+  cli_->set_read_timeout(std::chrono::seconds(1));
```
urfeex: At this point we should probably make the read timeout available as in the G5 implementation. the `GetConfiguredReceiveTimeout` method was so far only a dummy implementation. With these changes, I think it makes sense to keep a timeout member around. As far as I can see, cpp-httplib doesn't allow accessing the timeout, so we would have to keep track here.

We would also have to be sensitive about blocking calls such as `power_on` and `brake_release` that may take quite some time on real hardware. For that reason `power_on` has a timeout parameter already, but brake_release doesn't.

Other calls that might hit a timeout are

- loading programs (read timeout)
- importing (uploading) programs (write timeout)
- Updating programs (write timeout)
- Downloading programs (read timeout)

Using a higher default timeout (e.g. 10 seconds) might cover most of those cases already.

urfeex (changes requested):
Thanks a lot for this. However, I think to get this to a mergeable state, we would have to properly implement the concept of timeouts. The current `getConfiguredReceiveTimeout` is only a placeholder. We should fill that with life and probably also add a mechanism to alter the write timeout.

`power_on` would have to be updated to actually use the timeout parameter and reset to the default timeout afterwards, as done in G5.

@Vinicius-T-Robotics would you be up for extending this PR accordingly?

---

Edit: I see that while I was writing this, you have been addressing some of the points already :-)

Vinicius-T-Robotics:
> Thanks a lot for this. However, I think to get this to a mergeable state, we would have to properly implement the concept of timeouts. The current `getConfiguredReceiveTimeout` is only a placeholder. We should fill that with life and probably also add a mechanism to alter the write timeout.
> 
> `power_on` would have to be updated to actually use the timeout parameter and reset to the default timeout afterwards, as done in G5.
> 
> @Vinicius-T-Robotics would you be up for extending this PR accordingly?
> 
> Edit: I see that while I was writing this, you have been addressing some of the points already :-)

Hello, have added the write timeout as well in https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/518/changes/241c1a2787789b14462781962833c6f84a8b5491
Let me know what you think!

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -91,9 +156,36 @@ void DashboardClientImplX::disconnect()
...
+  {
+    tv = *recv_timeout_;
+  }
+  else
+  {
+    tv.tv_sec = 10;
+    tv.tv_usec = 0;
+  }
+  return tv;
+}
```
shafaq-trener, line 128: The initial tv.tv_sec = 1 is a dead store because both branches of the if/else overwrite it before it's ever read.

shafaq-trener (approved):
approved with a comment

include/ur_client_library/ur/dashboard_client_implementation_x.h
```diff
@@ -185,6 +186,11 @@ class DashboardClientImplX : public DashboardClientImpl
 
   std::unique_ptr<httplib::Client> cli_;
   VersionInformation robot_api_version_;
+  // Caller-configured timeouts. Null until the corresponding setter is called explicitly;
+  // the getters fall back to the constructor default (10 s) in that case. Mirrors the
+  // recv_timeout_ pattern used by DashboardClientImplG5.
+  std::unique_ptr<timeval> recv_timeout_;
+  std::unique_ptr<timeval> send_timeout_;
```
urfeex: How about not wrapping them into a unique_ptr and setting default values upon construction. Then, those can be used directly without checking for null.

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -59,28 +91,61 @@ std::string DashboardClientImplX::sendAndReceive([[maybe_unused]] const std::str
...
+  // call would otherwise reduce the in-flight deadline during the openapi.json GET.
+  timeval configured_tv = getConfiguredReceiveTimeout();
+  constexpr time_t kSetupMinSeconds = 10;
+  timeval setup_tv = configured_tv;
+  if (setup_tv.tv_sec < kSetupMinSeconds)
+  {
+    setup_tv.tv_sec = kSetupMinSeconds;
+    setup_tv.tv_usec = 0;
+  }
+  setReceiveTimeout(setup_tv);
```
urfeex: Did you create that on purpose? This section looks pretty much AO generated and tbh, I would doubt that fetching the `openapi.json` would take that long (I was around 200ms round trip time when testing a with a robot through a VPN over the internet).

This looks like an agent wanted to mimic what we've done in G5, which has other reasons.

urfeex (changes requested):

include/ur_client_library/ur/dashboard_client_implementation_x.h
Vinicius-T-Robotics: Good point. Removed https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/518/changes/cdd9988238ae31438e03fdde23df24be5c12f577

src/ur/dashboard_client_implementation_x.cpp
Vinicius-T-Robotics: This was probably triggered by my misunderstanding of this 
https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/518#discussion_r3395024344

Reverted in
https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/518/changes/e84c7693046fa556ea341eb11fa4fef4bca9c90a

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -49,6 +50,37 @@ DashboardClientImplX::DashboardClientImplX(const std::string& host) : DashboardC
...
+  // partition, paused container in tests).
+  //
+  // The 10 s default for read/write is chosen to cover blocking calls that legitimately take
+  // time on real hardware — brake_release, commandLoadProgram (read), commandUploadProgram
+  // (write), commandUpdateProgram (write), commandDownloadProgram (read) — without forcing
+  // each one to plumb its own timeout. commandPowerOn already accepts its own (longer)
+  // timeout parameter. Callers needing different limits can override via setReceiveTimeout.
+  cli_->set_connection_timeout(std::chrono::seconds(5));
+  cli_->set_read_timeout(std::chrono::seconds(10));
+  cli_->set_write_timeout(std::chrono::seconds(10));
```
urfeex:
```suggestion
  cli_->set_read_timeout(std::chrono::seconds(recv_timeout_.tv_sec) + std::chrono::microseconds(recv_timeout_.tv_usec));
  cli_->set_write_timeout(std::chrono::seconds(send_timeout_.tv_sec) + std::chrono::microseconds(send_timeout_.tv_usec));
```

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -61,7 +93,7 @@ bool DashboardClientImplX::connect([[maybe_unused]] const size_t max_num_tries,
 {
   std::string endpoint = base_url_ + "/openapi.json";
   // The PolyScope X Robot API doesn't require any connection prior to making calls. However, this
-  // check call will assure that the endpoint for making Robot API calls exist. This could fail if
+  // check call will assurea that the endpoint for making Robot API calls exist. This could fail if
```
urfeex:
```suggestion
  // check call will assure that the endpoint for making Robot API calls exist. This could fail if
```

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -91,10 +123,17 @@ void DashboardClientImplX::disconnect()
...
 timeval DashboardClientImplX::getConfiguredReceiveTimeout() const
 {
-  timeval tv;
-  tv.tv_sec = 1;
-  tv.tv_usec = 0;
-  return tv;
+  // If the caller has explicitly configured a receive timeout via setReceiveTimeout,
+  // return that. Otherwise fall back to the constructor default. See the constructor
+  // comment for the rationale on the 10 s default (covers brake_release / program
+  // load/upload/download without per-method timeouts).
```
urfeex:
```suggestion
```

src/ur/dashboard_client_implementation_x.cpp
```diff
@@ -91,10 +123,17 @@ void DashboardClientImplX::disconnect()
...
+  return recv_timeout_;
+}
+
+timeval DashboardClientImplX::getConfiguredSendTimeout() const
+{
+  // Mirrors getConfiguredReceiveTimeout. Default of 10 s matches the constructor.
```
urfeex:
```suggestion
```

urfeex (changes requested):
Adding a test setting timeouts to a couple of microseconds and validating that the calls then fail would be good. Apart from that, I think we're good to go.

Vinicius-T-Robotics:
> Adding a test setting timeouts to a couple of microseconds and validating that the calls then fail would be good. Apart from that, I think we're good to go.

Added a few tests for the timeouts. Note that timeout parameter resolution for httplib is actually milliseconds, so anything under 1000us seem to be rounded to zero.

urfeex (approved):
Thank you @Vinicius-T-Robotics !

PR #520, opened 2026-06-12
Send script blocking return refactor

### LINKED ISSUE #512: PrimaryClient: Improve error handling / reporting in sendScriptBlocking
### Feature summary

The new `sendScriptBlocking` method introduced in #425 currently returns a boolean success value.
Instead, it should inform callers on what went wrong.

In my personal opinion, the right interface for that would be `std::expected`, but since we don't do c++23 here, yet, throwing exceptions would be the most straightforward way.

Alternatively, a custom return type could be used, e.g.

```c++
struct ScriptExecutionResult
{
  bool success;
  // Those booleans could also be encoded into a REASON enum. They aren't necessarily all being checked.
  bool robot_mode_allows_execution;
  bool safety_mode_allows_execution;
  bool start_timeout;
  std::string message;
  std::optional<RuntimeExceptionMessage> runtime_execption;
  std::vector<ErrorCodeMessage> error_codes;
};
```

URJala:
The method now throws in any case where it would previously return false.

---

> [!NOTE]
> **High Risk**
> Breaking API change for all sendScriptBlocking callers; robot script execution error paths now throw instead of returning false, which can change control flow in integrators.
> 
> **Overview**
> **`PrimaryClient::sendScriptBlocking`** is now **`void`**: every former `false` outcome throws a typed **`UrException`** (e.g. **`TimeoutException`**, **`RobotModeException`**, **`SafetyModeException`**, **`RobotRuntimeException`**, **`RobotErrorCodeException`**, **`ReadOnlyInterfaceException`**, **`StreamNotConnectedException`**). Success is implied when the call returns normally.
> 
> Failure handling in **`sendScriptBlocking`** was reworked to build aggregated error messages, treat **`CRITICAL_FAULT`** like fault/violation during script execution, and map controller error **210** to **`ReadOnlyInterfaceException`**. Runtime failures are thrown as **`RobotRuntimeException`** with script context instead of only logging.
> 
> **`ReportLevel`** (plus **`reportLevelString`**) moved from the primary error-code message header into **`ur/datatypes.h`**, with **`CRITICAL_FAULT`** / **`DEVL_CRITICAL_FAULT`** added; **`PrimaryConsumer`** logging now covers those levels explicitly.
> 
> Docs, **`examples/send_script.cpp`**, and tests were updated for the exception-based API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db63a4f650dedf2596d0cd776552683f8a63738b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/send_script.cpp
+++ b/examples/send_script.cpp
@@ -45,13 +45,10 @@ def example_fun():
   relative_move = p[0,-0.1,0,0,0,0]
   movel(pose_trans(current_pose, relative_move), t=1)
 end)""";
-
-  if (client.sendScriptBlocking(fully_defined_script))
-  {
-    // The function definition can also be omitted
-    // A function name will then be auto generated
-    client.sendScriptBlocking(R"(textmsg("Successful program execution"))");
-  }
+  client.sendScriptBlocking(fully_defined_script);
+  // The function definition can also be omitted
+  // A function name will then be auto generated
+  client.sendScriptBlocking(R"(textmsg("Successful program execution"))");
   // A script-function name can also be passed to the method
   // A timeout can also be given to limit the wait for the passed function to start. If timeout = 0, it will
   // wait indefinitely.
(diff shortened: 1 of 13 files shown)


examples/send_script.cpp
```diff
@@ -66,18 +63,26 @@ end
...
+  }
+  catch (const RobotRuntimeException& exc)
+  {
+    URCL_LOG_INFO("Caught expected runtime exception from sendScriptBlocking");
+    URCL_LOG_INFO(exc.what());
+  }
+  catch (const UrException& exc)
+  {
+    URCL_LOG_INFO("Caught unexpected exception from sendScriptBlocking");
+    URCL_LOG_INFO(exc.what());
```
urfeex:
```suggestion
    URCL_LOG_ERROR("Caught unexpected exception from sendScriptBlocking");
    URCL_LOG_ERROR(exc.what());
```

urfeex (approved):

PR #528, opened 2026-06-24
PoC of streaming API for external_control

### LINKED ISSUE #306: Continuous execution of motion primitives with blending in InstructionExecuter when sending single primitives
### Feature summary

I would like to send new motion primitives live to the InstructionExecuter. Since the full trajectory is not necessarily known at the start, using a MotionSequence is not ideal. However, I still want to utilize blending between two primitives using blend_radius.
The issue is that the InstructionExecuter only processes the next primitive once the previous one has fully completed. This causes the motion to come to a stop between two primitives.
Is there a way to enable continuous execution of motion primitives with blending when sending single primitives instead of a MotionSequence?

### LINKED ISSUE #524: Support open-ended trajectories with URDriver
### Feature summary

Currently, when sending a trajectory with `URDriver::writeTrajectoryControlMessage(TRAJECTORY_START, ...)` you must provide the number of trajectory points that will be delivered up front, but the future is not always known. For more advanced usage, offer `writeTrajectoryControlMessage(TRAJECTORY_STREAM_START, -1, ...` to start a stream, and allow continuous delivery of trajectory points with `writeTrajectorySplinePoint` until a final `writeTrajectoryControlMessage(TRAJECTORY_STREAM_END` indicates that the stream is closed. The user of this API accepts the risks/complexity of ensuring that the arm does not starve for points and that it ends safely (e.g. 0/0 vel/accel).

acmorrow:
@urfeex - Here is what I had Claude put together to demonstrate the possibility of keeping open trajectories, per #524.

It includes:
- The addition of the `STREAM_START` and `STREAM_END` control messages to the `TrajectoryControlMesssage` enumeration.
- An integration test for streaming and an example program.
- Support for configuring the cmake build with a target IP for integration testing. Not perfect since it gets baked into the build, but not terrible either.
- Argument parsing fixes such that setting the robot IP on the command line works for various integration tests.
- Updates to the `external_contro.urscript` to support stream start and stream end.

I'm using this right now and it seems to work for some basic cases. I haven't stress tested it or thought deeply about the various edge cases or error scenarios that this may need to cover.

The goal here is really just to put some code behind the ideas over in #524 so we can discuss further. I'm happy to rework any of this as necessary.

I will also leave some comments directly on the review for a few things I want to call out.

---

> [!NOTE]
> **High Risk**
> Changes real-time robot motion control in `external_control.urscript` and the reverse-interface protocol; incorrect streaming semantics or edge-case bugs could truncate trajectories or report wrong completion results.
> 
> **Overview**
> Adds **open-ended trajectory streaming** alongside the existing finite `TRAJECTORY_START` path: producers open a stream with `TRAJECTORY_STREAM_START`, send motion primitives without a fixed upfront count, then close with `TRAJECTORY_STREAM_END` (and the same total count on `TRAJECTORY_CANCEL` in streaming mode).
> 
> **`external_control.urscript`** implements streaming with a `STREAMING_SENTINEL` point budget, `trajectory_streaming` state, guarded `STREAM_END`/cancel handling (so stray or duplicate end messages do not corrupt finite trajectories), and refined underrun vs. clean-drain behavior in `trajectoryThread`.
> 
> The C++ API documents the new `TrajectoryControlMessage` values on `ReverseInterface` / `UrDriver`. A new **`trajectory_streaming`** example and architecture/docs describe the handshake, producer pacing, and result codes. **Headless integration tests** cover success, underrun, cancel, and several regression cases for malformed streaming control sequences.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f98096d879c26f2fc72760439bfcfc4f21a303c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/doc/architecture/reverse_interface.rst
+++ b/doc/architecture/reverse_interface.rst
@@ -103,11 +103,13 @@ completes once they have been consumed. This is the mode used by
 A *streaming* trajectory is open-ended. The producer does not commit to a point count in advance; it
 opens the stream, writes points for as long as it likes, and signals end-of-stream explicitly. This
 suits points produced on the fly -- from a planner or a teleoperation source -- where the total length
-is unknown when motion begins. The message sequence is::
+is unknown when motion begins. The message sequence is:
+
+.. code-block:: c++
 
    writeTrajectoryControlMessage(TRAJECTORY_STREAM_START)   // open the stream
    // write a motion primitive, any number of times:
-   //   writeTrajectoryPoint(...) / writeTrajectorySplinePoint(...) / writeMotionPrimitive(...)
+   // writeTrajectoryPoint(...) / writeTrajectorySplinePoint(...) / writeMotionPrimitive(...)
    writeTrajectoryControlMessage(TRAJECTORY_STREAM_END, n)  // close the stream
 
 ``TRAJECTORY_STREAM_START`` opens the trajectory; its point-count argument is unused. The producer then
--- a/doc/architecture/reverse_interface.rst
+++ b/doc/architecture/reverse_interface.rst
@@ -49,8 +49,8 @@ meaning:
            - joint velocities (SPEEDJ)
            - trajectory instructions (FORWARD)
 
-             - field 1: Trajectory control mode(1: TRAJECTORY_MODE_RECEIVE, -1: TRAJECTORY_MODE_CANCEL)
-             - field 2: Number of trajectory points left to transfer
+             - field 1: Trajectory control mode (1: TRAJECTORY_MODE_RECEIVE, 2: TRAJECTORY_MODE_STREAM_START, 3: TRAJECTORY_MODE_STREAM_END, -1: TRAJECTORY_MODE_CANCEL). See :ref:`streaming_trajectories` for the streaming modes.
+             - field 2: Trajectory point count. Its interpretation depends on the control mode in field 1.
 
            - Cartesian velocities (SPEEDL)
            - Cartesian pose (POSE)
@@ -87,6 +87,63 @@ meaning:
    ``ReverseInterface`` class.
 
 Depending on the control mode one can use the ``write()`` (SERVOJ, SPEEDJ, SPEEDL, POSE, TORQUE), ``writeTrajectoryControlMessage()`` (FORWARD) or ``writeFreedriveControlMessage()`` (FREEDRIVE) function to write a message to the "reverse_socket".
+
+.. _streaming_trajectories:
+
+Streaming trajectories
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+A ``FORWARD``-mode trajectory can be executed in one of two ways.
+
+A *finite* trajectory declares its total length up front:
+``writeTrajectoryControlMessage(TRAJECTORY_START, n)`` promises exactly ``n`` points, and execution
+completes once they have been consumed. This is the mode used by
+:ref:`trajecotry_joint_interface_example`.
+
+A *streaming* trajectory is open-ended. The producer does not commit to a point count in advance; it
+opens the stream, writes points for as long as it likes, and signals end-of-stream explicitly. This
+suits points produced on the fly -- from a planner or a teleoperation source -- where the total length
+is unknown when motion begins. The message sequence is::
+
+   writeTrajectoryControlMessage(TRAJECTORY_STREAM_START)   // open the stream
+   // write a motion primitive, any number of times:
+   //   writeTrajectoryPoint(...) / writeTrajectorySplinePoint(...) / writeMotionPrimitive(...)
+   writeTrajectoryControlMessage(TRAJECTORY_STREAM_END, n)  // close the stream
+
+``TRAJECTORY_STREAM_START`` opens the trajectory; its point-count argument is unused. The producer then
+writes motion primitives -- via ``writeTrajectoryPoint()``, ``writeTrajectorySplinePoint()`` or
+``writeMotionPrimitive()`` -- for as long as it likes. ``TRAJECTORY_STREAM_END`` closes it.
+
+.. important::
+   The ``n`` passed to ``TRAJECTORY_STREAM_END`` must equal the total number of motion primitives the
+   producer wrote since ``TRAJECTORY_STREAM_START``. The controller consumes primitives asynchronously
+   as it executes them and cannot otherwise tell how many are still outstanding; it uses ``n`` to
+   account for those still in flight before completing. Aborting a stream with ``TRAJECTORY_CANCEL``
+   requires the same count, for the same reason. (For a finite ``TRAJECTORY_START`` trajectory the
+   cancel argument is unused.)
+
+Two obligations fall on the producer:
+
+- **Do not starve the controller.** The robot consumes the stream as it executes it, so the producer
+  must keep work available: the next primitive must arrive before the robot finishes the ones it
+  already holds. The budget for delivering it is simply the duration of the work in hand -- a primitive
+  that runs for five minutes gives five minutes to produce its successor, while a stream of
+  eight-millisecond primitives demands one every eight milliseconds. If the robot runs out of submitted
+  work before the next primitive arrives and before ``TRAJECTORY_STREAM_END`` is sent, the trajectory
+  ends with ``TRAJECTORY_RESULT_FAILURE``. Delivering primitives as they are produced, without holding
+  them back, is the natural way to satisfy this.
+
+- **End at rest.** Streaming adds no wind-down deceleration of its own. Make the final point a
+  controlled stop, with zero velocity and acceleration, so the spline interpolates smoothly to rest at
+  the goal. The trajectory thread issues ``stopj`` on exit regardless, so a non-zero terminal point
+  still stops the robot, but the transition will be less smooth.
+
+On completion the callback registered with ``setTrajectoryEndCallback()`` reports
+``TRAJECTORY_RESULT_SUCCESS`` for a clean end (``TRAJECTORY_STREAM_END`` processed, all points
+consumed), ``TRAJECTORY_RESULT_FAILURE`` for a producer underrun, or ``TRAJECTORY_RESULT_CANCELED`` if
+the stream was aborted with ``TRAJECTORY_CANCEL``.
+
+See :ref:`trajectory_streaming_example` for a complete, working example.
 
 .. _direct_torque_control_mode:
 
--- a/doc/examples/trajectory_streaming.rst
+++ b/doc/examples/trajectory_streaming.rst
@@ -1,0 +1,121 @@
+:github_url: https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/doc/examples/trajectory_streaming.rst
+
+.. _trajectory_streaming_example:
+
+Trajectory streaming example
+============================
+
+This example demonstrates open-ended *trajectory streaming*: a producer sends points to the
+controller without declaring a total point count up front, then signals end-of-stream when it is
+finished. Contrast this with the :ref:`trajecotry_joint_interface_example`, which uses the finite
+trajectory path where the point count is fixed in the initial ``TRAJECTORY_START`` message.
+
+The streaming communication contract -- the ``STREAM_START`` / ``STREAM_END`` handshake, the meaning
+of the end-of-stream point count, the producer's obligation not to starve the controller, and the
+trajectory result codes -- is described in :ref:`streaming_trajectories`. This page walks through the
+example code and refers back to that section for the authoritative contract.
+
+Setup
+-----
+
+As with the other examples, we create a ``UrDriver`` (here through ``ExampleRobotWrapper``) and
+register a callback that fires when a trajectory finishes:
+
+.. literalinclude:: ../../examples/trajectory_streaming.cpp
+   :language: c++
+   :caption: examples/trajectory_streaming.cpp
+   :linenos:
+   :lineno-match:
+   :start-at: const bool headless_mode = true;
+   :end-before: // --------------- PRE-POSITION VIA FINITE TRAJECTORY
+
+Trajectory execution is asynchronous, so completion is reported through a callback. The result
+distinguishes a successful run from a cancellation or a failure:
+
+.. literalinclude:: ../../examples/trajectory_streaming.cpp
+   :language: c++
+   :caption: examples/trajectory_streaming.cpp
+   :linenos:
+   :lineno-match:
+   :start-at: void trajDoneCallback
+   :end-at: }
+
+While a trajectory runs, the control PC must keep telling the robot program that the connection is
+still alive. We do this by pumping ``TRAJECTORY_NOOP`` messages while we wait; without them the robot
+program stops waiting for input and the ``external_control`` program exits:
+
+.. literalinclude:: ../../examples/trajectory_streaming.cpp
+   :language: c++
+   :caption: examples/trajectory_streaming.cpp
+   :linenos:
+   :lineno-match:
+   :start-at: // Pump TRAJECTORY_NOOP
+   :end-before: // Sample a quintic-Hermite
+
+Generating a motion
+-------------------
+
+To have something to stream, we sample a quintic-Hermite blend between two joint configurations. The
+blend has zero velocity and acceleration at both endpoints, which -- as the streaming contract
+recommends -- gives the final point the controlled-stop property needed to bring the robot cleanly to
+rest:
+
+.. literalinclude:: ../../examples/trajectory_streaming.cpp
+   :language: c++
+   :caption: examples/trajectory_streaming.cpp
+   :linenos:
+   :lineno-match:
+   :start-at: // Sample a quintic-Hermite
+   :end-before: int main(
+
+Pre-positioning
+---------------
+
+Before streaming, we park the robot at a known start pose using an ordinary finite trajectory. This
+doubles as a demonstration of the ``TRAJECTORY_START`` path for comparison:
+
+.. literalinclude:: ../../examples/trajectory_streaming.cpp
+   :language: c++
+   :caption: examples/trajectory_streaming.cpp
+   :linenos:
+   :lineno-match:
+   :start-after: // --------------- PRE-POSITION VIA FINITE TRAJECTORY
+   :end-before: // ----------------- STREAMING TRAJECTORY DEMO
+
+Streaming the trajectory
+------------------------
+
+We precompute all of the points up front. The final point inherits the ``qd = 0``, ``qdd = 0``
+boundary condition of the quintic blend, so it is a compliant controlled-stop terminal:
+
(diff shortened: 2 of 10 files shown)


examples/trajectory_streaming.cpp
```diff
@@ -0,0 +1,294 @@
...
+  {
+    if (!g_my_robot->getUrDriver()->writeTrajectorySplinePoint(p.q, p.qd, p.qdd, static_cast<float>(dt)))
+    {
+      URCL_LOG_ERROR("writeTrajectorySplinePoint failed mid-stream");
+      return 1;
+    }
+  }
+  const auto stream_end_send_wall = std::chrono::steady_clock::now();
+  g_my_robot->getUrDriver()->writeTrajectoryControlMessage(
+      urcl::control::TrajectoryControlMessage::TRAJECTORY_STREAM_END, k_num_points);
```
acmorrow, line 204: This is an important note: the client is obligated to provide the total number of points streamed when calling TRAJECTORY_STREAM_END, so that we can properly reset the counter so it counts down to zero and the trajectory ends naturally.

tests/CMakeLists.txt
```diff
@@ -25,11 +25,17 @@ include(GoogleTest)
 option(INTEGRATION_TESTS "Build the integration tests that require a running robot / URSim" OFF)
 option(POLYSCOPE_X_TESTS_WITH_REMOTE_CONTROL "Run Polyscope X tests that require remote control" OFF)
 option(CHECK_RTDE_DOCS_RECIPE "Fetch the RTDE documentation to auto-generate a recipe containing all output fields and check that with the RTDE client. Requires python3 and pandas and an internet connection." OFF)
+set(INTEGRATION_TESTS_ROBOT_IP "" CACHE STRING "If set, override the default robot IP (192.168.56.101) compiled into each integration test main() by passing --robot_ip <ip> to every ctest-registered integration test.")
```
acmorrow, line 28: I'm not sure why this is `CACHE` - it probably doesn't need to be. I'm also happy to find another mechanism to inject this state rather than a CMake variable. This was just an easy way to get it started. The really relevant change is computing `INTEGRATION_TESTS_ROBOT_IP_ARG` and then plumbing that through via `EXTRA_ARGS` on the test invocations

acmorrow:
@urfeex - I see the bot has flagged a few things, I'll take a look at those and post an update.

acmorrow:
@urfeex - Also, an update: today we got this running on a physical UR5e and we were able to stream to the arm. So, there is at least some evidence that this is feasible in the real world.

acmorrow:
@urfeex - Pushed an update to address the deficiencies that the cursor bot tagged. I had claude write some tests, we confirmed they failed, made fixes, confirmed they passed.

resources/external_control.urscript
```diff
@@ -1135,14 +1163,45 @@ while control_mode > MODE_STOPPED:
...
+        if trajectory_streaming:
+          # params_mult[3] is the total count of spline points the producer
+          # wrote to the trajectory socket since STREAM_START. The
+          # trajectoryThread has decremented trajectory_points_left once per
+          # consumed point starting from STREAMING_SENTINEL, so the consumed
+          # count is (STREAMING_SENTINEL - trajectory_points_left). The
+          # unread remainder still in the OS socket buffer is therefore
+          # the producer's total minus the consumer's consumed count.
+          trajectory_points_left = params_mult[3] - (STREAMING_SENTINEL - trajectory_points_left)
+          trajectory_streaming = False
```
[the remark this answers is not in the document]
acmorrow: Is that actually true? I suspect that this behavior is more or less the same as the non-streaming case.

include/ur_client_library/control/reverse_interface.h
```diff
@@ -120,15 +124,37 @@ class ReverseInterface
...
    *
-   * \param trajectory_action 1 if a trajectory is to be started, -1 if it should be stopped
-   * \param point_number The number of points of the trajectory to be executed
-   * \param robot_receive_timeout The read timeout configuration for the reverse socket running in the external
-   * control script on the robot. If you want to make the read function blocking then use RobotReceiveTimeout::off()
-   * function to create the RobotReceiveTimeout object
+   * \param trajectory_action One of the values of TrajectoryControlMessage. The value selects which
+   * trajectory-control action the URScript dispatcher takes, and dictates how \p point_number is
+   * interpreted:
+   * - TRAJECTORY_CANCEL (-1): Cancels the currently executing trajectory. \p point_number is unused.
```
urfeex: What happens to the remaining points if a trajectory is canceled? If I understand things correctly, the buffer cleanup will try to clean a very high number of points and then hit the socket read timeout as defined in `clearTrajectoryPointsThread()`. We could make this more effective (avoid the timeout) by passing the total number of points along as done with `TRAJECTORY_STREAM_END`. It should still work passing along 0 to not break behavior.

include/ur_client_library/control/reverse_interface.h
```diff
@@ -120,15 +124,37 @@ class ReverseInterface
...
+   * - TRAJECTORY_NOOP (0): No-op; serves as a keepalive on the reverse socket while a trajectory
+   *   executes. The producer must send these (or another trajectory-control message) periodically
+   *   while waiting for trajectory completion so that the URScript dispatcher's read does not
+   *   time out and exit the external_control program. \p point_number is unused.
+   * - TRAJECTORY_START (1): Begins a finite trajectory of declared length. \p point_number is the
+   *   total number of subsequent writeTrajectoryPoint / writeTrajectorySplinePoint / writeMotionPrimitive
+   *   calls the producer will issue on the trajectory socket. URScript reads exactly that many points
+   *   and then completes.
+   * - TRAJECTORY_STREAM_START (2): Begins an open-ended (streaming) trajectory. The producer streams
+   *   spline points and signals end-of-stream with TRAJECTORY_STREAM_END. \p point_number is unused.
```
urfeex: Is this restricted to splines only? From what I can see the same mechanism also works for sending `movej`, `movel`, etc.

resources/external_control.urscript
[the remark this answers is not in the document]
urfeex: I would say, it's the same. If the number of sent trajectory points is higher than the communicated number, trajectory points remain in the buffer after clearing. This is the case for both, a finite trajectory and a streaming trajectory.

urfeex (changes requested):
It would be nice if we could move the IP-address refactoring to a separate PR, so this one gets a bit smaller in scope.

From a first glance, this looks quite clean. We would need some documentation updates, both on the trajectory point interface and one for the example (which could contain a lot of the text written in the example itself right now).

include/ur_client_library/control/reverse_interface.h
urfeex: I just tested this and adding `trajectory_points_left = params_mult[3] - (STREAMING_SENTINEL - trajectory_points_left)` in the script code section for `elif params_mult[2] == TRAJECTORY_MODE_CANCEL:` made the cleanup function clean the remaining number of points not timing out on reading from the socket.

include/ur_client_library/control/reverse_interface.h
urfeex: I just tested this with the motion sequence from the instruction_executor example. This worked like a charm.

urfeex:
I've added the two issues that this PR will close when being merged.

acmorrow:
@urfeex - No problem about splitting out the testing changes, they were in their own commit anyway. I've made a new issue (#530) and a new PR (#531). I just made a new branch and cherry-picked out the relevant commit from this one.

We will need to sort out how to drop out that commit from this review since it is the base commit. Probably, we just review what is here until satisfied, and then I'll do a final rebase above master and force-push to drop out the now vacuous testing changes commit, but lmk if you want to handle it differently.

urfeex:
Yes, a rebase on master once the testing changes are merged should be the way to go.

urfeex:
@acmorrow I've merged #531, so you should be able to rebase this ontop of master.

acmorrow:
@urfeex - OK, rebased and force pushed.

include/ur_client_library/control/reverse_interface.h
acmorrow: @urfeex - I just want to confirm before I make the change: you suggest that for `writeTrajectoryControlMessage(TRAJECTORY_CANCEL, ...`, if the trajectory was started with `TRAJECTORY_STREAM_START`, then `point_number` becomes meaningful, and should equal the total number of points sent, just like it would be for `TRAJECTORY_STREAM_END`, and then around line 1196 in `external_control.urscript`, if we are in streaming mode we adjust the point count as indicated above?

include/ur_client_library/control/reverse_interface.h
acmorrow: That's great: it isn't a mechanism we happen to use right now, so not something I was thinking about testing. I could change that language from `spline points` to `points or primitives`?

include/ur_client_library/control/reverse_interface.h
urfeex: Yes, exactly. Do you think, that makes sense?

include/ur_client_library/control/reverse_interface.h
urfeex: How about `motion segments` or `motion primitives`?

acmorrow: Yes, I'll do `motion primitives` as that is what spline points actually reduce to, so it is I think the most accurate.

include/ur_client_library/control/reverse_interface.h
acmorrow: I think it makes total sense. I will get an update pushed.

acmorrow:
@urfeex - Pushed two commits: one with the pure comment change about splines, and another with the improvements around cancellation taking the number of points.

resources/external_control.urscript
```diff
@@ -1135,14 +1163,49 @@ while control_mode > MODE_STOPPED:
...
         request_trajectory_cleanup()
         stopj(STOPJ_ACCELERATION)
         wait_for_trajectory_cleanup()
+        # The streaming session, if any, has been forcibly ended; clear
+        # the flag so a subsequent stray STREAM_END is a no-op.
+	if trajectory_streaming:
+	  # In streaming mode, a cancel should come with the total count, just like a STREAM_END.
+	  trajectory_points_left = params_mult[3] - (STREAMING_SENTINEL - trajectory_points_left)
+          trajectory_streaming = False
+	end
```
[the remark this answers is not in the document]
acmorrow: Oops, good catch cursorbot, I should have read the surrounding cancel code more carefully. Will move this up.

acmorrow:
@urfeex - Just checking in on this, as I'm not aware of anything more you had asked to change in this review. But if I've overlooked a previous comment or you have additional thoughts, please let me know and I'll be happy to iterate on them.

urfeex:
No, I consider that to be in my field right now. This review is on my short list already, I will get to this before the end of the week!

include/ur_client_library/control/reverse_interface.h
```diff
@@ -48,9 +48,13 @@ namespace control
...
 {
-  TRAJECTORY_CANCEL = -1,  ///< Represents command to cancel currently active trajectory.
-  TRAJECTORY_NOOP = 0,     ///< Represents no new control command.
-  TRAJECTORY_START = 1,    ///< Represents command to start a new trajectory.
+  TRAJECTORY_CANCEL = -1,       ///< Represents command to cancel currently active trajectory.
+  TRAJECTORY_NOOP = 0,          ///< Represents no new control command.
+  TRAJECTORY_START = 1,         ///< Represents command to start a new finite trajectory of declared length.
+  TRAJECTORY_STREAM_START = 2,  ///< Represents command to start an open-ended streaming trajectory.
+  TRAJECTORY_STREAM_END = 3,    ///< Represents command to terminate an open-ended streaming trajectory.
+                                ///< The point_number argument must equal the total number of spline points
```
urfeex:
```suggestion
                                ///< The point_number argument must equal the total number of segments
```

resources/external_control.urscript
```diff
@@ -1135,11 +1163,46 @@ while control_mode > MODE_STOPPED:
...
       elif params_mult[2] == TRAJECTORY_MODE_CANCEL:
         textmsg("cancel received")
         kill thread_trajectory
+        # The streaming session, if any, has been forcibly ended; clear
+        # the flag so a subsequent stray STREAM_END is a no-op.
+	if trajectory_streaming:
+	  # In streaming mode, a cancel should come with the total count, just like a STREAM_END.
+	  trajectory_points_left = params_mult[3] - (STREAMING_SENTINEL - trajectory_points_left)
+          trajectory_streaming = False
+	end
```
urfeex: That sneaked in tabs for indentation. This probably works, since URScript isn't that sensitive to indentations, but at least it might render weirdly (as it does with GitHub). Suggesting to replace this with spaces, like the rest of the file:

```suggestion
        if trajectory_streaming:
          # In streaming mode, a cancel should come with the total count, just like a STREAM_END.
          trajectory_points_left = params_mult[3] - (STREAMING_SENTINEL - trajectory_points_left)
          trajectory_streaming = False
        end
```

urfeex (changes requested):
I have two cosmetic suggestions, but we are still missing the updated sphinx documentation.

Specifically, the reverse interface documentation in https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/doc/architecture/reverse_interface.rst needs a major update and should outline the communication contract as you've done in the example.

Also, we have accompanying sphinx pages for each example in https://github.com/UniversalRobots/Universal_Robots_Client_Library/tree/master/doc/examples, so that should be added, as well. As I have written above, this should contain most of the comments from the cpp file, so the cpp example gets a bit more compact to read on its own.

include/ur_client_library/control/reverse_interface.h
acmorrow: Applying this, but I may adjust to `motion primitives` to match the doc comment below.

resources/external_control.urscript
acmorrow: hah that's what I get for typing it myself rather than asking claude to do it. My editor is definitely not aware of urscript.

acmorrow:
@urfeex - I had claude put together the desired documentation. It got a little overzealous since my ambient instructions for my local development include running clang-format, and it decided that was something to do at the same time. Unfortunately, I don't know the exact version of clang-format you use, so it is possible that this isn't going to work out well.

I also asked claude to change around some of the existing language from the example, where it had gotten overly focused on sockets and buffers, and instead focus on the client obligation to keep the arm from starving for work. I think it reads better now, overall.

As always, happy for any feedback you have on this!

resources/external_control.urscript
```diff
@@ -1135,11 +1163,46 @@ while control_mode > MODE_STOPPED:
...
       elif params_mult[2] == TRAJECTORY_MODE_CANCEL:
         textmsg("cancel received")
         kill thread_trajectory
+        # The streaming session, if any, has been forcibly ended; clear
+        # the flag so a subsequent stray STREAM_END is a no-op.
+        if trajectory_streaming:
+          # In streaming mode, a cancel should come with the total count, just like a STREAM_END.
+          trajectory_points_left = params_mult[3] - (STREAMING_SENTINEL - trajectory_points_left)
+          trajectory_streaming = False
+        end
```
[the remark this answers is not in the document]
acmorrow: @cursoragent - Isn't that a violation of the documented client obligations? They are required to pass a correct point count when cancelling a trajectory.

examples/trajectory_streaming.cpp
urfeex: This is now well-covered in the documentation, so I'm resolving this.

resources/external_control.urscript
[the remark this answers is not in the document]
urfeex: @acmorrow I agree.

In general, I think I would like to go to a proper buffer in script code instead of using the network stream as a buffer, but that's for a later change.

doc/architecture/reverse_interface.rst
```diff
@@ -88,6 +88,63 @@ meaning:
...
+``writeTrajectoryControlMessage(TRAJECTORY_START, n)`` promises exactly ``n`` points, and execution
+completes once they have been consumed. This is the mode used by
+:ref:`trajecotry_joint_interface_example`.
+
+A *streaming* trajectory is open-ended. The producer does not commit to a point count in advance; it
+opens the stream, writes points for as long as it likes, and signals end-of-stream explicitly. This
+suits points produced on the fly -- from a planner or a teleoperation source -- where the total length
+is unknown when motion begins. The message sequence is::
+
+   writeTrajectoryControlMessage(TRAJECTORY_STREAM_START)   // open the stream
```
urfeex:
```suggestion
is unknown when motion begins. The message sequence is:

.. code-block:: c++

   writeTrajectoryControlMessage(TRAJECTORY_STREAM_START)   // open the stream
```

doc/architecture/reverse_interface.rst
```diff
@@ -88,6 +88,63 @@ meaning:
...
+   // write a motion primitive, any number of times:
+   //   writeTrajectoryPoint(...) / writeTrajectorySplinePoint(...) / writeMotionPrimitive(...)
```
urfeex:
```suggestion
   // writeTrajectoryPoint(...) / writeTrajectorySplinePoint(...) / writeMotionPrimitive(...)
```

doc/examples/trajectory_streaming.rst
```diff
@@ -0,0 +1,121 @@
...
+.. literalinclude:: ../../examples/trajectory_streaming.cpp
+   :language: c++
+   :caption: examples/trajectory_streaming.cpp
+   :linenos:
+   :lineno-match:
+   :start-at: void trajDoneCallback
+   :end-at: }
+
+While a trajectory runs, the control PC must keep telling the robot program that the connection is
+still alive. We do this by pumping ``TRAJECTORY_NOOP`` messages while we wait; without them the robot
```
urfeex:
```suggestion
still alive. We do this by sending ``TRAJECTORY_NOOP`` messages while we wait; without them the robot
```

urfeex (approved):
I have only a couple of small suggestions. If the tests succeed, I can apply them and merge.

Thank you @acmorrow for putting all the effort into this!

acmorrow:
> I have only a couple of small suggestions. If the tests succeed, I can apply them and merge.
> 

Yes, suggestions all reasonable and I see you committed them. I do see some failed checks but I think you are better positioned to decide what they mean.

> Thank you @acmorrow for putting all the effort into this!

No problem! We are really excited about being able to do this, and very happy that it turned out to be within reach on such a short timeline. Thanks for all the review feedback and helping to land it.

urfeex:
Everything I expect to succeed, succeeds. Happily merging this :-)

PR #529, opened 2026-06-25
Fix RTDE control_step

urrsk:
Correct RTDE register name to control_step
Is available from Polyscope version 5.26.0 and 10.13.0


urfeex (approved):
Verified against 10.13.0

urfeex:
I will merge this once the upstream documentation is updated. Otherwise every integration test job will fail once this is merged.

PR #531, opened 2026-06-26
enable running integration tests against host networked ursim

### LINKED ISSUE #530: Support configuring the build to test against a different robot IP
### Feature summary

The robot isn't always on 192.168.56.101. It only requires some minor changes to enable configuring the build so that the tests as invoked by ctest run with a different IP.

acmorrow:
Fixes #530. This was pulled out from #528, as requested.

---

> [!NOTE]
> **Low Risk**
> Test and CMake-only changes; no production library or runtime behavior is modified.
> 
> **Overview**
> Integration tests can target a robot or URSim at an address other than the compiled-in default **`192.168.56.101`**, which supports host-networked URSim (e.g. **`127.0.0.1`**).
> 
> CMake adds **`INTEGRATION_TESTS_ROBOT_IP`**. When set at configure time, every ctest-registered integration test receives **`--robot_ip <ip>`** via **`gtest_add_tests` `EXTRA_ARGS`**, alongside existing flags like **`--headless`**.
> 
> Test **`main()`** handlers no longer **`break`** after the first recognized flag; they **`++i`** past flag values so **`--robot_ip`** and **`--headless`** can both apply when ctest passes multiple arguments.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69eb4019567e6ea51899ee5ae9c33acf6f083bbe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -25,11 +25,17 @@ include(GoogleTest)
 option(INTEGRATION_TESTS "Build the integration tests that require a running robot / URSim" OFF)
 option(POLYSCOPE_X_TESTS_WITH_REMOTE_CONTROL "Run Polyscope X tests that require remote control" OFF)
 option(CHECK_RTDE_DOCS_RECIPE "Fetch the RTDE documentation to auto-generate a recipe containing all output fields and check that with the RTDE client. Requires python3 and pandas and an internet connection." OFF)
+set(INTEGRATION_TESTS_ROBOT_IP "" CACHE STRING "If set, override the default robot IP (192.168.56.101) compiled into each integration test main() by passing --robot_ip <ip> to every ctest-registered integration test.")
 # Build Tests
 if (INTEGRATION_TESTS)
-  # Integration tests require a robot reachable at 192.168.56.101. Therefore, they have to be
-  # activated separately.
-  
+  # Integration tests require a reachable robot / URSim. The compiled-in default
+  # address is 192.168.56.101; configure with -DINTEGRATION_TESTS_ROBOT_IP=<ip>
+  # to override (e.g. 127.0.0.1 for a host-networked URSim).
+  set(INTEGRATION_TESTS_ROBOT_IP_ARG "")
+  if(INTEGRATION_TESTS_ROBOT_IP)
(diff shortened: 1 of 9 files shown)


tests/CMakeLists.txt
```diff
@@ -25,11 +25,17 @@ include(GoogleTest)
 option(INTEGRATION_TESTS "Build the integration tests that require a running robot / URSim" OFF)
 option(POLYSCOPE_X_TESTS_WITH_REMOTE_CONTROL "Run Polyscope X tests that require remote control" OFF)
 option(CHECK_RTDE_DOCS_RECIPE "Fetch the RTDE documentation to auto-generate a recipe containing all output fields and check that with the RTDE client. Requires python3 and pandas and an internet connection." OFF)
+set(INTEGRATION_TESTS_ROBOT_IP "" CACHE STRING "If set, override the default robot IP (192.168.56.101) compiled into each integration test main() by passing --robot_ip <ip> to every ctest-registered integration test.")
```
acmorrow, line 28: As I mentioned in the other review, let me know if you would prefer to inject this some other way. Also not sure why claude felt it should be `CACHE`: I've always been a little fuzzy on the whole CMake cache thing.

urfeex, line 28: From my CMake understanding, setting a cache variable should be the right way to do it.

urfeex (approved):

urfeex:
One thing that comes to my mind is that using the loopback device might not work with all tests, as it may interfere with 

- tests where we start fake servers on the test machine that use the same ports as the UR interfaces, e.g. [here](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/0b3d1e31372f0d21496e7b6da476cb78f06f50d6/tests/test_rtde_client.cpp#L796)
- tests where we use the loopback device as a known IP that we can connect to while we don't expect to have robot interfaces available, e.g. [here](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/0b3d1e31372f0d21496e7b6da476cb78f06f50d6/tests/test_rtde_client.cpp#L631-L633).

Thus, I would not expect that all tests succeed when using the loopback IP. This doesn't make this PR any less legitimate, that's just a comment, as the OP contained `127.0.0.1` as an explicit example.

PR #537, opened 2026-07-10
Use tcp_nodelay on socket server

urfeex:
The sockets we use in this library should be optimized for real-time communication, not for throughput. Therefore, they should use tcp_nodelay by default.

---

> [!NOTE]
> **Low Risk**
> Low-latency TCP tuning and non-fatal logging around setsockopt; no auth or data-model changes.
> 
> **Overview**
> **TCP sockets are tuned for lower latency** by enabling `TCP_NODELAY` on each **accepted** server connection (client outbound paths already set it in `TCPSocket::setupOptions`).
> 
> Socket option setup is centralized in **`setSocketOptionAndWarnOnError`**, which logs a warning when `setsockopt` fails instead of ignoring errors. **Listen socket** options (`SO_REUSEADDR`, `SO_KEEPALIVE`) and **client** options (`TCP_NODELAY`, `TCP_QUICKACK`, `SO_RCVTIMEO`) now go through that helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21bab872535f72a89b303e61b7dfb70b65130def. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/comm/tcp_server.cpp
+++ b/src/comm/tcp_server.cpp
@@ -261,8 +261,11 @@ void TCPServer::handleConnect()
     }
   }
 
-  int flag = 1;
-  ur_setsockopt(client_fd, IPPROTO_TCP, TCP_NODELAY, &flag, sizeof(int));
+  if (accepted)
+  {
+    constexpr int flag = 1;
+    setSocketOptionAndWarnOnError(client_fd, IPPROTO_TCP, TCP_NODELAY, &flag, sizeof(flag), "TCP_NODELAY");
+  }
 
   {
     std::lock_guard<std::mutex> lk(callback_mutex_);
--- a/src/comm/tcp_server.cpp
+++ b/src/comm/tcp_server.cpp
@@ -26,7 +26,9 @@
  */
 //----------------------------------------------------------------------
 
-#include <netinet/tcp.h>
+#ifndef _WIN32
+#  include <netinet/tcp.h>
+#endif
 #include <ur_client_library/log.h>
 #include <ur_client_library/comm/tcp_server.h>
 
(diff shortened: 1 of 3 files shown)


src/comm/tcp_server.cpp
```diff
@@ -257,6 +260,10 @@ void TCPServer::handleConnect()
       ur_close(client_fd);
     }
   }
+
+  int flag = 1;
+  ur_setsockopt(client_fd, IPPROTO_TCP, TCP_NODELAY, &flag, sizeof(int));
```
michal-milkowski: 1. check return value. Warn if option can't be set.
2. it's better to use sizeof(flag) instead.
3. this could be attempted on closed socket on line 260. either check if accepted == true or move flag set somewhere up

src/comm/tcp_server.cpp
```diff
@@ -257,6 +260,10 @@ void TCPServer::handleConnect()
       ur_close(client_fd);
     }
   }
+
+  int flag = 1;
```
michal-milkowski: can be constexpr

```suggestion
  constexpr int flag = 1;
```

michal-milkowski (changes requested):
Is nodelay also set on client socket?

src/comm/tcp_server.cpp
urfeex: Thank you for the comments. I implemented those in c73d273

urfeex:
> Is nodelay also set on client socket?

Yes, it is:
https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/b4371167cc0f298dbba1d8f9e960dcf3799284a8/src/comm/tcp_socket.cpp#L56

I've updated that to use the `setSocketOptionAndWarnOnError` function, as well.

michal-milkowski (approved):

PR #544, opened 2026-08-03
Throw RTDEInvalidKeyException on unknown RTDE key

urfeex:
This allows parsing information about failed fields. Since RTDEInvalidKeyException inherits from UrException, this is not API-breaking.

This should help in generate an explicit user output in https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1551

---

> [!NOTE]
> **Low Risk**
> Narrow change to error typing and exception payload on RTDE recipe validation; still inherits from UrException so existing catch-by-base behavior largely holds, with clearer structured data for new handlers.
> 
> **Overview**
> When RTDE output setup fails because recipe keys are unknown to the robot (and unavailable outputs are not ignored), the client now throws **`RTDEInvalidKeyException`** instead of a generic **`UrException`**, carrying the list of bad keys and the existing error text.
> 
> **`RTDEInvalidKeyException`** gains a constructor that accepts **`invalid_keys`** plus optional message text, and exposes **`invalid_keys`** as a public field so callers (e.g. ROS2 driver) can report which fields failed without parsing **`what()`**.
> 
> Tests for invalid input/output recipe keys use GMock **`ThrowsMessage`** / **`HasSubstr`**, link **`rtde_tests`** against **`gmock_main`**, and add **`TestableRTDEClient`** to exercise **`init()`** with an injected bad output recipe.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 544404489720a802afae2fa360935a9e2e27e820. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/include/ur_client_library/exceptions.h
+++ b/include/ur_client_library/exceptions.h
@@ -286,12 +286,20 @@ public:
   {
   }
 
+  explicit RTDEInvalidKeyException(const std::vector<std::string>& invalid_keys, const std::string& text = "")
+    : std::runtime_error(text)
+  {
+    this->invalid_keys = invalid_keys;
+  }
+
   virtual ~RTDEInvalidKeyException() = default;
 
   virtual const char* what() const noexcept override
   {
     return std::runtime_error::what();
   }
+
+  std::vector<std::string> invalid_keys;
 };
 
 class RTDEInputConflictException : public UrException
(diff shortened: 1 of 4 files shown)


urrsk (changes requested):
Suggest adding a test covering this improvement

urrsk (approved):

PR #546, opened 2026-08-04
Harden script_reader against vulnerabilities

urfeex:
This doesn't change anything in behavior, it merely enforces the behavior as it is described in the docs.

1. Path traversal (.. and absolute-path escape) — `replaceIncludes()` now rejects absolute include paths outright (avoiding the `path::operator/` LHS-drop trap) and requires every include's `weakly_canonical`-normalized target to be lexically inside 
`root_dir_` (the directory of the top-level script).

2. Unbounded / circular includes → stack overflow — added `include_stack_` (set of canonical paths currently being processed) for cycle detection and `include_depth_` bounded by a fixed threshold of 32. Both self-references and A→B→A cycles now throw `UrException`, as do chains beyond the depth cap.

3. `script_path_` state corruption from nested includes — removed the mutable `script_path_` member. The public `readScriptFile()` is now a thin wrapper that initializes state, and a new private `readScriptFileImpl()` save-restores `current_dir_` (and cleans up `include_stack_`/`include_depth_`) around every recursive include, even on exceptions.

4. Malformed include regex ['|"] — changed to ['"]. {% include |target.txt| %} is no longer treated as a valid include.

---

> [!NOTE]
> **High Risk**
> Changes security-sensitive filesystem include resolution for robot control scripts; legitimate deep include chains beyond 32 levels will now fail, and malicious or malformed include paths are rejected instead of read.
> 
> **Overview**
> **Hardens `ScriptReader` include handling** so nested `{% include %}` matches documented behavior and blocks path escape, unbounded recursion, and corrupted relative resolution.
> 
> `readScriptFile` now establishes a **script root** (`root_dir_`), tracks **current directory** per nesting level via private `readScriptFileImpl`, and replaces the mutable `script_path_` with `include_stack_` / `include_depth_` (cap **32**). Includes reject absolute and rooted paths, resolve targets with `weakly_canonical`, and require the result to stay under the root; cycles and depth overflow throw `UrException`. `filesystem_error` from canonicalization is wrapped with context for the main script path and include paths.
> 
> The include regex is fixed from `['|"]` to `['"]` so `|` is no longer treated as a quote. **Security regression tests** cover traversal, absolute paths, cycles, depth limits, sibling includes, Windows-style rooted paths, pipe-as-quote, and symlink canonicalization errors (POSIX).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52c3f42371272eb53cd3e6b2d7fa872699c195e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/include/ur_client_library/control/script_reader.h
+++ b/include/ur_client_library/control/script_reader.h
@@ -29,9 +29,11 @@
 // -- END LICENSE BLOCK ------------------------------------------------
 
 #pragma once
+#include <cstddef>
 #include <filesystem>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 #include <variant>
 
 #include <ur_client_library/ur/datatypes.h>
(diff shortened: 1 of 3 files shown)


urrsk (changes requested):
For this one, it would be good to have full test coverage, and it seems, at first glance, that the few missing parts are possible to do with not too much effort.

urrsk (approved):

Issue #550, opened 2026-08-11
Clean recovery from starving the arm during streaming is currently not possible


acmorrow:
### Affected Client Library version(s)

2.14.0 2.14.1

### What combination of platform is the Client Library running on.

Mac

### How did you install the Client Library

Build the library from source

### Which robot platform is the library connected to.

URSim in docker

### Robot SW / URSim version(s)

master

### How are you using the Client Library

Headless without using the teach pendant

### Issue details

# Summary
With `external_control.urscript`, If a streaming client fails to meet its obligation to keep the arm fed with points and allows the arm to "run off the end of the stream", the script is left in a state that makes safe recovery impossible. There is not a clear workaround at this time.

## Issue details
The contract for streaming is that the client will end the stream with a point count, so that `external_control.urscript` knows how many more points to read off the socket. That works well, both for normal stream termination and for client-initiated cancellation. However, if the client fails to meet its obligation to keep the arm fed with points, the trajectory will report failure. But without a terminal message from the client containing the total point count, the `external_control.urscript` script can't know how many points may be sitting in the buffer that it should drain. Nor can it guess, or rely on a timeout, since the client may already be sending points for some future trajectory.

## Steps to Reproduce
Run the arm off the end of a trajectory. I have written a number of unit tests which I can share on a branch that make the problem clear.

## Expected Behavior
After a trajectory failure, the script should be able to safely drain any points already in the socket buffer.

## Actual Behavior
The points remain in the buffer, and dangerously become part of the next trajectory.

# Workaround Suggestion
I do not currently have a workaround suggestion. The most obvious approach would be to ask streaming clients to follow up a trajectory failure with an explicit STREAM_END or CANCEL containing a total point count. But that has some problems, in that it is likely to trigger a second "completion" on the same trajectory: confusing for both the script and the client. That could probably be suppressed in the `external_control.urscript` logic, but now things are starting to get a little messy. Another possibility would be to require streaming clients to attach an ID to each streamed trajectory point, which would allow the arm side script to drain points with matching ids, but this is imperfect as once you have read the first non-matching point you've already done the wrong thing. Another option would be to introduce a requirement on clients that when they get failure during a stream, they must drop and reconnect, but that's harsh and graceless.

Basically, I'd like to use this issue to talk through the issues. I have some unit tests that demonstrate the problem, please let me know if you would like me to publish them to a branch.

[...]

PR #552, opened 2026-08-11
Add tests demonstrating failure to recover from stream starvation

acmorrow:
Hi @urrsk -

This PR, which I'm opening as a draft because some of the tests are currently failing, demonstrates the failure to correctly recover to a valid state after starving the arm of points during a streaming move.

There are also some new tests that currently pass, acting as controls/validation for existing behavior. They came into being as I was validating my understanding of the urscript behavior. But they seem like worthwhile tests on their own, and useful to prevent regressions. Easy enough to drop them if you just want the failing tests.

---

> [!NOTE]
> **High Risk**
> Breaking wire/API change (3.x) requires library and bundled URScript to upgrade together; changes real-time trajectory execution and recovery paths on the robot.
> 
> **Overview**
> Introduces a **per-move identifier** on the trajectory/reverse-socket protocol so the robot can ignore stale trajectory points after a stream fails, is cancelled, or a producer keeps writing briefly. Trajectory records grow from 21 to **22** int32 fields; trajectory start messages on the reverse socket now carry **`move_id`** (formerly padding). **`UrDriver`** increments and assigns IDs on `TRAJECTORY_START` / `TRAJECTORY_STREAM_START`; **`TrajectoryPointInterface::setMoveId()`** stamps each outgoing point. **`writeTrajectoryControlMessage()`** gains a required **`move_id`** argument (documented as a **3.x** breaking change in migration notes).
> 
> On the robot, **`external_control.urscript`** tracks `trajectory_move_id`, skips mismatched points in `trajectoryThread`, drains full records in cleanup, and on **mid-stream underrun** clears `trajectory_streaming` and `trajectory_points_left` so the next move is not blocked by `STREAMING_SENTINEL` cleanup (issue **#550**).
> 
> Adds unit tests for wire encoding of move IDs and broad **integration tests** for streaming gaps, underrun recovery (finite and stream), stale points after underrun, and cancel-with-understated count—with RTDE joint checks where motion must be verified.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9196723c2d9881e8db8e2b41ef99ee6ff7cf74db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_trajectory_streaming.cpp
```diff
@@ -383,6 +383,290 @@ TEST_F(TrajectoryStreamingTest, stream_end_with_overcount_yields_failure)
...
+  for (int i = 0; i < k_num_points; ++i)
+  {
+    ASSERT_TRUE(
+        g_my_robot->getUrDriver()->writeTrajectorySplinePoint(i % 2 == 0 ? pose_b : pose_a, zero, zero, k_step_time));
+  }
+
+  // Stop writing to the trajectory socket for 500ms while the thread consumes
+  // what we already sent. We keep answering the reverse interface with NOOPs so
+  // that only the trajectory socket goes quiet.
+  for (int i = 0; i < 10; ++i)
```
urrsk: Maybe sleep slightly shorter than 0.5 sec as this is exactly the timeout define on the robot, to avoid unstable test and the time what is used to compare here might run on different computers.

acmorrow:
@urrsk - I've pushed a possible solution to cleanup after starvation based on adding an id to points in a move. I followed up on your comment in #550 to discuss some of the ways my approach differs from your notes. Definitely interested to see what you think of this. It does seem to make the newly introduced tests go green, so that's a good start.

tests/test_trajectory_streaming.cpp
acmorrow: Sure, that's reasonable. I'll cut it to 400ms.

include/ur_client_library/control/reverse_interface.h
```diff
@@ -159,9 +166,9 @@ class ReverseInterface
    * \see examples/trajectory_point_interface.cpp for a finite-trajectory usage example, or
    * examples/trajectory_streaming.cpp for an open-ended streaming trajectory example.
    */
-  bool
-  writeTrajectoryControlMessage(const TrajectoryControlMessage trajectory_action, const int point_number = 0,
-                                const RobotReceiveTimeout& robot_receive_timeout = RobotReceiveTimeout::millisec(200));
+  bool writeTrajectoryControlMessage(
+      const TrajectoryControlMessage trajectory_action, const int32_t move_id, const int point_number,
```
acmorrow: @urrsk - I'm not sure why claude decided to opt for `int32_t` here rather than following along with the other parameters declared as `int`. I think consistency is better, so I'll be changing the `move_id` parameter here to `int`.

tests/test_trajectory_streaming.cpp
acmorrow: Done

include/ur_client_library/control/reverse_interface.h
acmorrow: Done

acmorrow:
Hi @urrsk - Just checking in on this PR. Generally, does it seem like a viable approach? Can the wire protocol change be accepted? Please let me know if there is anything you need from me in order to help move a fix for this issue forward, whether under this mechanism or some other that you would prefer. I'm happy to rework it if needed.

urfeex:
Without having dived into the details too much, it is obvious that this PR would require a major version bump. We do have a couple of things in mind that we want to change with the next major bump, so it might be reasonable to start preparing a 3.x on a new branch where this could potentially get merged to. Would you agree @urrsk ?

acmorrow:
> Without having dived into the details too much, it is obvious that this PR would require a major version bump. We do have a couple of things in mind that we want to change with the next major bump, so it might be reasonable to start preparing a 3.x on a new branch where this could potentially get merged to. Would you agree @urrsk ?

@urfeex that plan makes sense to me. For now, we will probably work against this PR for our stuff since it fixes a real use case for us.

doc/migration_notes.rst
```diff
@@ -5,6 +5,23 @@ This document contains notes on the migration of the ur_client_library between m
 
 It contains only breaking changes.
 
+Migrating from 2.x.x to 3.x.x
```
acmorrow, line 8: @urfeex - Wanted to call this to your attention, since I note that you tagged this into the 2.15 version. This was speculative that the major version would bump with the wire, so if that isn't true, this should be reverted.

urfeex: We'll make a decision today when to include this. I've merely added it to the milestone so that I don't make a release without considering this beforehand. Most likely, this will indeed go to a 3.x release, though.

urfeex: I've moved it to a 2.15.1 milestone. We think that we can do this without an API breaking change (I will make a review about this later).

The wire-breaking changes are acceptable, since we consider the communication between the script code and the library an internal asset. We'll add a clear statement about this, separately. We have been doing wire-breaking changes in the past, as well.

include/ur_client_library/ur/ur_driver.h
```diff
@@ -1136,6 +1137,13 @@ class UrDriver
   std::unique_ptr<control::ScriptSender> script_sender_;
   std::unique_ptr<control::ScriptReader> script_reader_;
 
+  // Identifies the move that is currently being sent to the robot. The driver is what assigns these
+  // identifiers, because it is the only object which observes both the boundaries between moves and
+  // every point written within them. It issues one identifier for each command that begins a move,
+  // and never reuses one for the life of a connection. A value of 0 belongs to no move, and is what
+  // the robot holds before it has been told about any.
+  std::atomic<int32_t> trajectory_move_id_ = { 0 };
```
urfeex, line 1145: Could we move this to the ReverseInterface, instead? This way we would not need to expose it on the ReverseInterface's `writeTrajectoryControlMessage()` at all and remove the API break of this PR.

doc/architecture/reverse_interface.rst
```diff
@@ -51,6 +51,7 @@ meaning:
 
              - field 1: Trajectory control mode (1: TRAJECTORY_MODE_RECEIVE, 2: TRAJECTORY_MODE_STREAM_START, 3: TRAJECTORY_MODE_STREAM_END, -1: TRAJECTORY_MODE_CANCEL). See :ref:`streaming_trajectories` for the streaming modes.
              - field 2: Trajectory point count. Its interpretation depends on the control mode in field 1.
+             - field 3: The identifier of the move that is being started, for the two control modes which start one. The robot adopts this identifier and then executes only those trajectory points which carry it, and that is what prevents points left over from a move that has already ended from running as part of the next one. The remaining control modes do not start a move, and the robot ignores the field for them. See :ref:`trajectory_point_interface` for the point record which carries the identifier back.
```
urfeex, line 54: One way to avoid wire-breaking here, would be to re-use the `ScriptReader::isVariableRegistered(const std::string& key)` function introduced in 01292bdf4ebb7e45a436c8930042e0001ed7e92c.

urfeex (changes requested):
If we avoid doing API breakages here, we can merge this onto 2.x. Please also remove the migration text in that case.

doc/architecture/reverse_interface.rst
acmorrow: Convenient timing on that! I think that looks straightforward enough. I'll put something together and add it to the PR.

doc/migration_notes.rst
acmorrow: OK, that would be great if we can get it done without a API break. I'll leave this comment here as a reminder to back this out, but I'd like to continue the discussion about how exactly to achieve the API-compatible version on your other comment in `ur_driver.h`.

include/ur_client_library/ur/ur_driver.h
acmorrow: The thing that first gave me some pause about relocating the `move_id_` counter into `ReverseInterface` is that `TrajectoryPointInterface` is-a `ReverseInterface` (as is `ScriptCommandInterface`). So suddenly everyone gets a `move_id_`, whether they want one or not. But that's already true of other fields in `ReverseInterface` that aren't relevant to the other types.

Then I realized that `TrajectoryPointInterface` already declares its own copy of the move id. So, sort of a happy accident, but we can just use the inherited one for that. I'm not particularly keen on inherited protected mutable state, but, again, it isn't too terribly divergent from what is already happening amongst these types.

Also, having thought about it more, I don't think any of the move_id stuff needs to use atomics. The obligation w.r.t. thread-safety in this library falls, as far as I can tell, to the caller. Specifically, if they are interacting via multiple threads, they are required to ensure the happens-before / happens-after ordering around starting/ending trajectories and enqueuing points. So I'm inclined to change it to a non-atomic, independently of where it ultimately lives. I think making it atomic suggests to a reader thread-safety behavior that isn't globally supported in the library. Your thoughts on that idea would be appreciated.

acmorrow:
@urfeex - Just letting you know I haven't forgotten about this or your PR comments, I've just been busy with something else. I should be able to get back to this early next week.

PR #556, opened 2026-08-19
Fix flaky spline interpolation test

URJala:
Made new method: sendTrajectoryAndConfirmStart which launches an async thread to confirm that a new trajectory resets the spline_travel_time and then starts executing. This is more robust than the previous implementation.

I have replaced all the calls to sendTrajectory where they were immediately followed by waitForTrajectoryStarted, with one call to sendTrajectoryAndConfirmStart. waitForTrajectoryStarted is no longer used, but I havent removed it yet.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to spline interpolation integration test harness code; no production library behavior is modified.
> 
> **Overview**
> Addresses flaky spline interpolation integration tests by **synchronizing trajectory send with RTDE-based start detection** instead of polling after the fact.
> 
> **`sendTrajectoryAndConfirmStart`** starts an async **`confirmTrajectoryStarted`** loop before sending points. That thread watches `output_double_register_1` for a spline travel-time reset (new transfer) and then for travel time increasing (execution started), while the main thread sends the trajectory once the helper is ready. Shared state uses **`std::atomic`** for `g_trajectory_running` and the async flags, plus mutex/condition-variable handshaking and 1s timeouts on failure.
> 
> **`sendTrajectory`** now passes **`RobotReceiveTimeout::off()`** on **`TRAJECTORY_START`**. Tests that previously called **`sendTrajectory`** then **`waitForTrajectoryStarted`** (and manual **`g_trajectory_running`**) now **`ASSERT_TRUE(sendTrajectoryAndConfirmStart(...))`**; **`waitForTrajectoryStarted`** remains but is unused in those paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1fdafd6d5619c9398830ab7939661012d4d5527. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_spline_interpolation.cpp
+++ b/tests/test_spline_interpolation.cpp
@@ -171,7 +171,7 @@ protected:
 
     // Send trajectory to robot for execution
     ASSERT_TRUE(g_my_robot->getUrDriver()->writeTrajectoryControlMessage(
-        urcl::control::TrajectoryControlMessage::TRAJECTORY_START, s_pos.size()));
+        urcl::control::TrajectoryControlMessage::TRAJECTORY_START, s_pos.size(), RobotReceiveTimeout::off()));
     for (size_t i = 0; i < s_pos.size(); i++)
     {
       // QUINTIC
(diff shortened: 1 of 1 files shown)


tests/test_spline_interpolation.cpp
```diff
@@ -254,21 +291,114 @@ class SplineInterpolationTest : public ::testing::Test
...
+          if (!g_my_robot->getUrDriver()->writeTrajectoryControlMessage(
+                  urcl::control::TrajectoryControlMessage::TRAJECTORY_NOOP))
+          {
+            std::cout << "Failed to write trajectory NOOP control message, exiting confirm thread" << std::endl;
+            return false;
+          }
+        }
+
+        // Confirm that the new trajectory is running
+        if (std::abs(spline_travel_time) > start_spline_travel_time)
```
urfeex, line 367: Wouldn't this be problematic when there was a long trajectory running beforehand? I don't remember by hard when we reset the spline_travel_time, but it seems fragile to rely on it being small in the beginning. Once it has been reset, being larger than a reasonable threshold should be sufficient, right?

tests/test_spline_interpolation.cpp
```diff
@@ -381,12 +518,10 @@ TEST_F(SplineInterpolationTest, cubic_spline_with_end_point_velocity)
...
 
   // Send the trajectory to the robot
-  sendTrajectory(s_pos, s_vel, std::vector<urcl::vector6d_t>(), s_time);
-  g_trajectory_running = true;
-
   // Make sure that the trajectory has started before we start testing for trajectory execution
-  waitForTrajectoryStarted();
+  ASSERT_TRUE(sendTrajectoryAndConfirmStart(s_pos, s_vel, std::vector<urcl::vector6d_t>(), s_time));
 
+  g_trajectory_running = true;
```
urfeex: We could probably set this inside `sendTrajectoryAndConfirmStart`, right?

urfeex (changes requested):

tests/test_spline_interpolation.cpp
```diff
@@ -254,21 +291,114 @@ class SplineInterpolationTest : public ::testing::Test
...
+    }
+
+    // Allow new trajectory to be sent
+    {
+      std::lock_guard<std::mutex> lk(async_mutex);
+      async_ready = true;
+      async_cv.notify_one();
+    }
+
+    bool spline_travel_time_reset = false;
```
urfeex:
```suggestion
    bool spline_travel_time_reset = start_spline_travel_time == 0.0;
```

tests/test_spline_interpolation.cpp
URJala: It only gets to this stage if we have seen that it is either zero, or at least has become smaller (where start_spline_travel_time is reset), and then we check that it advances one timestep.

tests/test_spline_interpolation.cpp
URJala: Yes, I think we could

urfeex (approved):

PR #559, opened 2026-08-26
 Fix spline point velocity and acceleration precision

s4lmon:
The spline vel/acc are quantised at `1e-6` by `MULT_JOINTSTATE`, so near-zero accelerations reconstruct as jagged profiles. On our UR15 we notice many current compensation errors, that make the arm unusable with the ROS 2 driver. 

I think this is a problem only encountered by the larger arms, given higher intertia.

We have this fix that is working, implementing the same strategy as #482.

We implement a finer dedicated multiplier (`MULT_VEL_ACC = 1e8`) and an explicit range guard (~21.47 rad/s(²)), analagous to max goal time check. 

This is backwards compatible. We achieve this by only setting the range guard if the modern encoding is used. `UrDriver` detects whether the control script defines the new multiplier via `{{VEL_ACC_REPLACE}}`. If undefined, we fall back to legacy `MULT_JOINTSTATE` encoding with a warning.

Credit to @tdivic for discovering this bug and implementing a fix in our ROS 2 driver.

---

> [!NOTE]
> **Medium Risk**
> Changes real-time trajectory wire encoding and robot script decoding for splines; mismatch between library and script would corrupt motion, though auto-detection and legacy fallback mitigate custom-script risk.
> 
> **Overview**
> Spline trajectory points now encode **velocities and accelerations** with a dedicated **`MULT_VEL_ACC` (1e8)** instead of **`MULT_JOINTSTATE` (1e6)**, so near-zero values survive the int32 wire format and the robot no longer reconstructs jagged acceleration profiles that can fault the controller.
> 
> **`TrajectoryPointInterface`** applies the finer scale only to spline second/third blocks (positions stay on joint-state scale), rejects spline vel/acc above ~21.47 rad/s(²), and exposes **`setVelAccMultiplier` / `getVelAccMultiplier`**. The bundled **`external_control.urscript`** adds **`MULT_velacc`** via **`{{VEL_ACC_REPLACE}}`** and decodes cubic/quintic spline **`qd`/`qdd`** with that multiplier.
> 
> **Backward compatibility:** **`ScriptReader`** records which placeholders were actually substituted (**`isVariableRegistered`**). On init, **`UrDriver`** uses fine encoding when the script contains **`VEL_ACC_REPLACE`**; otherwise it logs a warning and falls back to legacy **`MULT_JOINTSTATE`** encoding so older custom scripts keep working. Architecture docs and tests cover the new multiplier, range checks, near-zero roundtrip, and driver init behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e05eb2bfe2b82120e0820e2088025743b2d6db5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/doc/architecture/trajectory_point_interface.rst
+++ b/doc/architecture/trajectory_point_interface.rst
@@ -57,11 +57,11 @@ representations in 21 datafields. The data fields have the following meaning:
           - For all MOVEC variants this field contains the via point (same
             joint-vs-pose interpretation as the target at indices 0-5, see the motion type at
             index 20).
-          - trajectory point velocities (multiplied by ``MULT_JOINTSTATE``) for spline joint types
+          - trajectory point velocities (multiplied by ``MULT_VEL_ACC``) for spline joint types
 
    12-17  Depending on the motion type, this represents either
 
-          - trajectory point accelerations (multiplied by ``MULT_JOINTSTATE``) for spline joint
+          - trajectory point accelerations (multiplied by ``MULT_VEL_ACC``) for spline joint
             types.
 
           - for all other motion types
@@ -112,11 +112,19 @@ where
 
 - ``MULT_JOINTSTATE``: 1000000
 - ``MULT_TIME``: 1000000
+- ``MULT_VEL_ACC``: 100000000
 
 .. note::
    With ``MULT_TIME`` being 1000000, the maximum duration that can be sent is 2147 seconds, while
    precision is cut off at 1 microsecond. (The same applies to the blend radius, respectively being
    max 2147 m and 1 μm precision.)
+
+.. note::
+   Spline point velocities and accelerations use the finer ``MULT_VEL_ACC`` scaling, giving 1e-8
+   resolution with a maximum magnitude of ~21.47 rad/s (rad/s²). The coarser ``MULT_JOINTSTATE``
+   scaling quantizes near-zero accelerations so heavily that the reconstructed acceleration
+   profile becomes jagged, which can trigger controller faults such as "Compensation current
+   disagreement". Values exceeding the maximum magnitude are rejected by the library.
 
 .. note::
    The ``*_POSE`` / ``*_JOINT`` motion-type variants let callers mix joint-space and Cartesian
--- a/doc/architecture/trajectory_point_interface.rst
+++ b/doc/architecture/trajectory_point_interface.rst
@@ -122,9 +122,8 @@ where
 .. note::
    Spline point velocities and accelerations use the finer ``MULT_VEL_ACC`` scaling, giving 1e-8
    resolution with a maximum magnitude of ~21.47 rad/s (rad/s²). The coarser ``MULT_JOINTSTATE``
-   scaling quantizes near-zero accelerations so heavily that the reconstructed acceleration
-   profile becomes jagged, which can trigger controller faults such as "Compensation current
-   disagreement". Values exceeding the maximum magnitude are rejected by the library.
+   scaling quantizes near-zero accelerations, causing reconstructed acceleration
+   profile to become jagged, which can trigger controller faults. Values exceeding the maximum magnitude are rejected by the library.
 
 .. note::
    The ``*_POSE`` / ``*_JOINT`` motion-type variants let callers mix joint-space and Cartesian
--- a/doc/architecture/trajectory_point_interface.rst
+++ b/doc/architecture/trajectory_point_interface.rst
@@ -125,6 +125,10 @@ where
    scaling quantizes near-zero accelerations, causing reconstructed acceleration
    profile to become jagged, which can trigger controller faults. Values exceeding the maximum magnitude are rejected by the library.
 
+   Finer scaling is only used when the control script defines the matching multiplier through
+   the ``{{VEL_ACC_REPLACE}}`` placeholder. For user-supplied scripts without that placeholder the
+   library falls back to encoding spline velocities and accelerations with ``MULT_JOINTSTATE``.
+
 .. note::
    The ``*_POSE`` / ``*_JOINT`` motion-type variants let callers mix joint-space and Cartesian
    targets freely through the high-level APIs (see :ref:`instruction_executor` and the
--- a/src/control/script_reader.cpp
+++ b/src/control/script_reader.cpp
@@ -179,6 +179,11 @@ std::filesystem::path canonicalizeOrThrow(const std::filesystem::path& input, co
 }
 }  // namespace
 
+bool ScriptReader::wasVariableUsed(const std::string& key) const
+{
+  return used_variables_.count(key) != 0;
+}
+
 std::string ScriptReader::readScriptFile(const std::string& filename, const DataDict& data)
 {
   // Top-level entry point: normalize the input path, set up the include root, reset the include
@@ -190,6 +195,7 @@ std::string ScriptReader::readScriptFile(const std::string& filename, const Data
   current_dir_.clear();
   include_stack_.clear();
   include_depth_ = 0;
+  used_variables_.clear();
 
   return readScriptFileImpl(canonical_input, data);
 }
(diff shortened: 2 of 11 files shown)


src/control/trajectory_point_interface.cpp
```diff
@@ -180,24 +199,18 @@ bool TrajectoryPointInterface::writeMotionPrimitive(const std::shared_ptr<contro
...
+    {
+      buffer[index++] = htobe32(static_cast<int32_t>(round(item * mult)));
+    }
+  };
+
+  // Only spline points carry per-sample vel/acc in the second and third block.
+  const int32_t vel_acc_mult = primitive->type == MotionType::SPLINE ? MULT_VEL_ACC : MULT_JOINTSTATE;
+  write_block(first_block, MULT_JOINTSTATE);
+  write_block(second_block, vel_acc_mult);
+  write_block(third_block, vel_acc_mult);
```
s4lmon, line 219: prefer this but happy to revert for readability

doc/architecture/trajectory_point_interface.rst
```diff
@@ -57,11 +57,11 @@ representations in 21 datafields. The data fields have the following meaning:
           - For all MOVEC variants this field contains the via point (same
             joint-vs-pose interpretation as the target at indices 0-5, see the motion type at
             index 20).
-          - trajectory point velocities (multiplied by ``MULT_JOINTSTATE``) for spline joint types
+          - trajectory point velocities (multiplied by ``MULT_VEL_ACC``) for spline joint types
```
s4lmon, line 60: name ok?

urrsk:
Hi @s4lmon 
Thanks for pointing this out and your proposal.

src/ur/ur_driver.cpp
```diff
@@ -111,6 +112,7 @@ void UrDriver::init(const UrDriverConfiguration& config)
   control::ScriptReader::DataDict data;
   data[JOINT_STATE_REPLACE] = std::to_string(control::ReverseInterface::MULT_JOINTSTATE);
   data[TIME_REPLACE] = std::to_string(control::TrajectoryPointInterface::MULT_TIME);
+  data[VEL_ACC_REPLACE] = std::to_string(control::TrajectoryPointInterface::MULT_VEL_ACC);
```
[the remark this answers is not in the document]
s4lmon: I've fixed it, so it's backwards compatible and defaults to the finer resolution.

I can't find any code `TrajectoryPointInterface` directly (bypassing `UrDriver`) but if anyone were doing this, it would now default to the finer encoding and must call `setVelAccMultiplier(MULT_JOINTSTATE)` when paired with legacy scripts.

tests/test_trajectory_point_interface.cpp
```diff
@@ -525,12 +525,12 @@ TEST_F(TrajectoryPointInterfaceTest, write_splines_accelerations)
...
-  EXPECT_EQ(send_vel[2], ((double)received_velocities[2]) / traj_point_interface_->MULT_JOINTSTATE);
-  EXPECT_EQ(send_vel[3], ((double)received_velocities[3]) / traj_point_interface_->MULT_JOINTSTATE);
-  EXPECT_EQ(send_vel[4], ((double)received_velocities[4]) / traj_point_interface_->MULT_JOINTSTATE);
-  EXPECT_EQ(send_vel[5], ((double)received_velocities[5]) / traj_point_interface_->MULT_JOINTSTATE);
+  EXPECT_EQ(send_vel[0], ((double)received_velocities[0]) / traj_point_interface_->MULT_VEL_ACC);
+  EXPECT_EQ(send_vel[1], ((double)received_velocities[1]) / traj_point_interface_->MULT_VEL_ACC);
+  EXPECT_EQ(send_vel[2], ((double)received_velocities[2]) / traj_point_interface_->MULT_VEL_ACC);
+  EXPECT_EQ(send_vel[3], ((double)received_velocities[3]) / traj_point_interface_->MULT_VEL_ACC);
+  EXPECT_EQ(send_vel[4], ((double)received_velocities[4]) / traj_point_interface_->MULT_VEL_ACC);
+  EXPECT_EQ(send_vel[5], ((double)received_velocities[5]) / traj_point_interface_->MULT_VEL_ACC);
```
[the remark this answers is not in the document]
s4lmon: Fixed.

doc/architecture/trajectory_point_interface.rst
urfeex: that seems fine.

src/control/script_reader.cpp
```diff
@@ -220,7 +226,7 @@ std::string ScriptReader::readScriptFileImpl(const std::filesystem::path& canoni
   try
   {
     script_code = readFileContent(path_key);
-    replaceVariables(script_code, data);
+    replaceVariables(script_code, data, variable_registry_);
     replaceConditionals(script_code, data);
```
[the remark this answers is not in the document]
urfeex: Since the conditionals can also take variables into account, they have to parsed afterwards. I think, it's fair to do the check as done in this PR.

src/control/trajectory_point_interface.cpp
```diff
@@ -97,6 +99,24 @@ bool TrajectoryPointInterface::writeMotionPrimitive(const std::shared_ptr<contro
...
+    auto spline_primitive = std::static_pointer_cast<control::SplinePrimitive>(primitive);
+    auto within_range = [max_vel_acc](const vector6d_t& values) {
+      return std::all_of(values.begin(), values.end(),
+                         [max_vel_acc](const double value) { return std::abs(value) <= max_vel_acc; });
+    };
+    if (!within_range(spline_primitive->target_velocities) ||
+        (spline_primitive->target_accelerations.has_value() &&
+         !within_range(spline_primitive->target_accelerations.value())))
+    {
+      URCL_LOG_ERROR("Spline point velocity or acceleration out of range. Maximum allowed magnitude is %f.",
```
urfeex: Printing the actual value might be helpful to narrow it down.

src/control/trajectory_point_interface.cpp
urfeex: I'm happy with that.

src/control/trajectory_point_interface.cpp
```diff
@@ -328,6 +342,11 @@ void TrajectoryPointInterface::messageCallback([[maybe_unused]] const socket_t f
   }
 }
 
+void TrajectoryPointInterface::setVelAccMultiplier(const int32_t multiplier)
+{
+  mult_vel_acc_ = multiplier;
+}
```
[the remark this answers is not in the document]
urfeex: I would agree with that.

urfeex (changes requested):
Could you please add a test for a legacy script that it falls back to using MULT_JOINTSTATE?

src/control/trajectory_point_interface.cpp
s4lmon: Good call. Fixed

s4lmon:
> Could you please add a test for a legacy script that it falls back to using MULT_JOINTSTATE?

Done

src/control/trajectory_point_interface.cpp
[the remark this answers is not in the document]
s4lmon: Done

s4lmon:
@urfeex one thing I omitted: 

I echoed the fix you made earlier for timing with resolution. On our own arms, we have been post-processing trajectories before sending them and clip near-zero velocities and accelerations to zero. Of course, the fault isn't caused by actual lost precision, but by the quantisation hitting near-zero values inconsistently, where some accelerations will round to zero while neighbours will not, and this is what causes the jagged profiles. 

We don't have insight into how these are processed on the closed-source side of the UR stack so this is a proposed solution here. If you prefer, I can implement velocity clipping on values below 1e-4 rad/s(^2), which we have running locally. The advantage of this is also that their is no wire-format change and will work unchanged with existing scripts. I did not pursue this fix, however, as it avoids such an arbitrary constant which would still lead to some technical quantisation error (of 1e-4).

If you prefer the above solution let me know.

urfeex:
> If you prefer the above solution let me know.

I think, the solution proposed in this PR is fine.

urfeex (approved):
this looks good to me.

PR #565, opened 2026-09-09
Tolerate null values in the PolyScope X program list

mairtirol:
## What

`DashboardClientImplX::commandGetProgramList()` reads each program's fields straight
out of the JSON response and into the `ProgramInformation` constructor. Any field that
comes back as `null` is therefore converted directly into a `std::string` or an
`unsigned int`, which nlohmann rejects:

```
[json.exception.type_error.302] type must be string, but is null
```

The conversion happens inside the loop, so a single program with a null field fails the
whole query — including `name`, which is the only field many callers need.

Only `lastModifiedDate` and `lastSavedDate` were guarded. This PR guards the remaining
nullable fields the same way.

## Why null is a valid response

The Robot API schema (`/universal-robots/robot-api/openapi.json`) declares five of the
six `ProgramInformation` fields as nullable:

```json
"required": ["name", "description"],
"properties": {
  "name":             {"type": "string"},
  "description":      {"anyOf": [{"type": "string"},  {"type": "null"}]},
  "createdDate":      {"anyOf": [{"type": "integer"}, {"type": "null"}]},
  "lastSavedDate":    {"anyOf": [{"type": "integer"}, {"type": "null"}]},
  "lastModifiedDate": {"anyOf": [{"type": "integer"}, {"type": "null"}]},
  "programState":     {"anyOf": [{"$ref": "#/components/schemas/ProgramState-Output"},
                                 {"type": "null"}]}
}
```

`name` is the only field declared non-nullable, and it is the only one still read
directly.

## How to reproduce

On a real robot (firmware 10.13.1.0, Robot API 3.1.7.0):

1. Export a program to a USB stick from the teach pendant.
2. Import it again from the USB stick.
3. `GET /universal-robots/robot-api/programs/v1/`

The imported program comes back with a null description, even though the exported
`.urpx` contains one:

```json
{"name":"my_program","description":null,"createdDate":1788882381574,
 "lastSavedDate":null,"lastModifiedDate":null,"programState":"FINAL"}
```

`commandGetProgramList()` then throws `type_error.302` and no program list can be
retrieved at all. For us this makes the UR backend unable to complete its
initialisation against a real controller, since querying the program list is the first
thing it does.

The exported `.urpx` does contain a description, so the import path appears to drop it.
That may be a bug in PolyScope X rather than in this library.

Two details that may be useful:

- A program *created* on the teach pendant gets `""`, not `null` — only the USB
  import path produces the null.

## Verification
Built against a real robot with a null-description program installed:

Present in 2.14.1, 2.15.0 and current `master`.

---

> [!NOTE]
> **Low Risk**
> Localized JSON parsing change in program list handling with regression tests; no auth or persistence impact.
> 
> **Overview**
> Fixes **`commandGetProgramList()`** so PolyScope X program list responses no longer fail when the Robot API returns **`null`** for schema-nullable fields (`createdDate`, `description`, `programState`, plus the date fields that were already guarded).
> 
> Nullable values are now read with **`is_null()`** checks and mapped to **empty strings or `0`** before building **`ProgramInformation`**, instead of letting nlohmann throw **`type_error.302`** and abort parsing of the entire **`programs`** array. **`name`** remains the only field read without a null guard.
> 
> Adds **mock HTTP tests** for fully populated programs, null **`description`**, all nullable fields null, mixed lists where one entry has nulls, and an empty list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9be8728c6a26426171f440822d318e321fcbceb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/ur/dashboard_client_implementation_x.cpp
+++ b/src/ur/dashboard_client_implementation_x.cpp
@@ -715,6 +715,16 @@ DashboardResponse DashboardClientImplX::commandGetProgramList()
     std::vector<ProgramInformation> programs;
     for (auto prog : json_data["programs"])
     {
+      // The Robot API schema declares every field of ProgramInformation except
+      // "name" as nullable, so each of them has to be read defensively. Reading
+      // a null directly into the constructor's std::string / unsigned int
+      // parameters throws a json type_error and fails the whole query.
+      unsigned int created = 0;
+      if (!prog["createdDate"].is_null())
+      {
+        created = static_cast<unsigned int>(prog["createdDate"]);
+      }
+
       unsigned int last_modified = 0;
       if (!prog["lastModifiedDate"].is_null())
       {
@@ -727,8 +737,19 @@ DashboardResponse DashboardClientImplX::commandGetProgramList()
         last_saved = static_cast<unsigned int>(prog["lastSavedDate"]);
       }
 
-      ProgramInformation pi(prog["createdDate"], prog["description"], last_modified, last_saved, prog["name"],
-                            prog["programState"]);
+      std::string description;
+      if (!prog["description"].is_null())
+      {
+        description = prog["description"];
+      }
+
+      std::string program_state;
+      if (!prog["programState"].is_null())
+      {
+        program_state = prog["programState"];
+      }
+
+      ProgramInformation pi(created, description, last_modified, last_saved, prog["name"], program_state);
       programs.push_back(pi);
     }
     response.data["programs"] = programs;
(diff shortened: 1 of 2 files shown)


urfeex (commented):
This does look fine, thank you for addressing this. I would like to add unittests covering this, before approving and merging. @mairtirol if you want to help on that, this will probably speed up things. Otherwise I will try to look into adding tests.

mairtirol:
@urfeex I tried my best :-). I added tests in 9be8728 - five cases in the existing DashboardClientImplXMockTest fixture (httplib server, no robot) covering the nullable ProgramInformation fields.

urfeex (approved):
This is looking good.

Coverage is shown as partial for the null-check lines, since the branch where `prog` is not a JSON object would throw. This would only happen if the robot sends us invalid JSON or we would not put the json response into the response object correctly, in which case all tests on that endpoint would probably fail. Therefore, it should be fine to not cover that branch.

PR #571, opened 2026-09-14
Add workflow for creating releases

urfeex:
This workflow automatically creates a GitHub release when a release-PR is merged.


urrsk (changes requested):
Looking at the bot's feedback, it seems the tag might be troublesome and maybe the draft is what we should do in the first place?

urrsk (approved):

PR #576, opened 2026-09-17
Filter RTDE output tests by documented PolyScope versions

urrsk:


urfeex:
Does this make sense in light of #560?

PR #577, opened 2026-09-18
Add support for g-series robots

urfeex:
- RobotType definitions
- RobotSeries definition
- Startupscript arguments (preparation for 10.15.0)


urrsk (approved):

urrsk (changes requested):
It seems all our integration test now are running the RTDEClientTest.check_rtde_data_fields_match_docs test which Failed.
It is understandable that they are failing, but it should only be the "Latest" that runs it, right!

urrsk:
> It seems all our integration test now are running the RTDEClientTest.check_rtde_data_fields_match_docs test which Failed. It is understandable that they are failing, but it should only be the "Latest" that runs it, right!

My bad. It is of cause failing but have nothing to do with this code

urrsk (approved):

